PR #1, opened 2022-07-25
Initial implementation of nodejs client

mshustov:
The initial implementation of nodejs client based on [Language clients spec](https://docs.google.com/document/d/19zAlnKvMgu075_DDA_k2rY03D_P4wrH6RwBJKJGa8zE/edit#heading) and [Language spec. One pager](https://docs.google.com/document/d/1e0tMosUUFln0o4PzXXwumOa_YFcmR5wDj6ARxPG-tFY/edit).

## Supported features
- inserting data in a row-based form.
  - inserting an array or a stream of JSON objects in `JSONCompactEachRow ` format. We can consider adding support for other formats. Still, IMO we should minimize the number of supported formats in the `insert` operation if we want to convert data into `Native` or `RowBinary` format. We can discuss it later once we have performance testing to compare the throughput of different formats. It didn't tune any streaming settings until we have performance benchmarks to compare different options.
  - inserting a string or a stream of strings is not supported yet. 
- selecting data. see a list of supported formats in https://github.com/ClickHouse/clickhouse-js/pull/1/files#diff-8a7fc60079b51ce7aba6f7e75c059e5b52d7f1eb162c5827dae1ac4d84d2b267
  - as a string
  - as a JSON object
  - a stream of strings or JSON objects.
- sending a command to execute. No response is returned.
- ping a server
- parametrized queries. The query parameters binding is delegated to the ClickHouse.
- instance- and query-level settings
- creates a pool of connections with [Http.Agent](https://nodejs.org/api/http.html#new-agentoptions)
- parsing of ClickHouse errors
- primitive logging will be extended in the future

## High-level overview of the architecture

`ClickHouse API and Transport` are implemented in a single `ClickHouseClient` class.
`ClickHouseClient` - exposes protocol-agnostic API to communicate to the ClickHouse server. It manages query formatting, data coding/encoding via `DataFormatter` and creating connections via `Connection` factory.

`Cluster manager` and `Connection manager` are not implemented in the MVP phase since we do not support connections to multiple hosts. These components are presented on the diagram for reference. 

Network communication is implemented via the `HttpAdapter` class, which hides an implementation of a specific network protocol. `HttpAdapter` creates a pool of open sockets to a **single host** (uses the nodejs default - `Infinity` number of open sockets). Since the client supports connection to a single host, host health tracking isn't implemented.
`ClickHouseClient` doesn't work with `HttpAdapter` directly. It expects every adapter to implement the `Connection` interface instead.
```typescript
interface Connection {
  ping(): Promise<boolean>;
  close(): Promise<void>;
  select(params: SelectParams): Promise<Stream.Readable>
  command(params: CommandParams): Promise<void>
  insert(params: InsertParams): Promise<void>
}
```

`Data formatter`  restructures language-specific data structures from ClickHouse output / to ClickHouse input formats.

## Client API
The client API surface consists of 4 methods:
### Ping
```typescript
ping: () => Promise<boolean>
```

### Command
```typescript
interface CommandParams {
  query: string;
  clickhouse_settings?: ClickHouseSettings;
  query_params?: Record<string, unknown>;
}
command: (params: CommandParams) => Promise<void>;
```

### Insert
```typescript
interface InsertParams {
  table: string;
  values : ReadonlyArray<any> | Stream.Readable;
  clickhouse_settings?: ClickHouseSettings;
  query_params?: Record<string, unknown>;
}
insert: (params: InsertParams) => Promise<void>;
```

### Select
```typescript
interface SelectParams {
  query: string;
  format?: DataFormat;
  clickhouse_settings?: ClickHouseSettings;
  query_params?: Record<string, unknown>;
}
select: (params: SelectParams) => Promise<Rows>;
```
The result of `select` operation is `Rows` object. I added it to simplify support of different output formats: 
- string
- JSON
- Stream
   - of strings
   - of objects

The interface and functionality of the `Rows` class are discussable and probably don't make much sense for some languages (go, for example).
I'm going to extend it with `summary`, `column names`, `column types`, and `query id` later.
```typescript
interface Rows {
  text: () => Promise<string>;
  json: () => Promise<any>;
  asStream: () => ReadableStream of Row
  
  // not implemented yet
  summary: Summary;
  columnName: string[];
  columnTypes: DataTypes[];
  queryId: string;
}

interface Row {
  text: () => string;
  json: () => any;
}
```

### Demo
see integration tests in `__tests__/integration` folder: https://github.com/ClickHouse/clickhouse-js/tree/fb58b9497f755ab8288aaae93a740995ed503a6f/__tests__/integration

### DIFF
--- a/__tests__/integration/command.test.ts
+++ b/__tests__/integration/command.test.ts
@@ -45,7 +45,7 @@ describe('command', () => {
     }
   });
 
-  it.skip('can specify a parametrized query', async () => {
+  it.skip('can specify a parameterized query', async () => {
     client = createClient();
     await client.command({
       query: 'CREATE TABLE {table_name: String} (id UInt64, name String, sku Array(UInt8), timestamp DateTime) Engine = Memory',
--- a/__tests__/integration/insert.test.ts
+++ b/__tests__/integration/insert.test.ts
@@ -38,7 +38,7 @@ describe('insert', () => {
     ]);
   });
 
-  it('can insert strings with non-latin symols', async() => {
+  it('can insert strings with non-latin symbols', async() => {
     const dataToInsert = [
       [42, "привет", [0,1]],
       [43, "мир", [3,4]]
--- a/__tests__/integration/select.test.ts
+++ b/__tests__/integration/select.test.ts
@@ -38,7 +38,7 @@ describe('select', () => {
     expect(response).toBe('0\n1\n');
   });
 
-  it('can specify a parametrized query', async () => {
+  it('can specify a parameterized query', async () => {
     client = createClient();
     const rows = await client.select({
       query: 'SELECT number FROM system.numbers WHERE number > {min_limit: UInt64} LIMIT 3',
@@ -190,7 +190,7 @@ describe('select', () => {
   });
 
   describe('select result asStream()', () => {
-    it('trows an exception if format is not stream-able', async () => {
+    it('throws an exception if format is not stream-able', async () => {
       expect.assertions(1);
       client = createClient();
       const result = await client.select({
--- a/src/client.ts
+++ b/src/client.ts
@@ -152,9 +152,9 @@ export class ClickHouseClient {
     return await this.connection.close();
   }
 }
-
-const formatRe = /\bformat\b/i;
-function validateSelectQuery(query: string): void {
+// /i.test('select format from table format json limit 1')
+const formatRe = /\bformat\b\s([a-z]*)$/i;
+export function validateSelectQuery(query: string): void {
   if(formatRe.test(query)){
     throw new Error('Specifying format is not supported, use "format" parameter instead.')
   }
--- a/src/data_formatter/formatter.ts
+++ b/src/data_formatter/formatter.ts
@@ -8,7 +8,7 @@ const supportedFormats = [
   'JSONCompactStringsEachRowWithNames',
   'JSONCompactStringsEachRowWithNamesAndTypes',
   'CSV',
-  'TabSeparated'
+  'TabSeparated',
 ] as const;
 export type DataFormat = typeof supportedFormats[number];
 
@@ -27,7 +27,7 @@ type StreamableJsonDataFormat = typeof streamableJSONFormats[number];
 const streamableFormat = [
   ...streamableJSONFormats,
   'CSV',
-  'TabSeparated'
+  'TabSeparated',
 ] as const;
 type StreamableDataFormat = typeof streamableFormat[number];
 
--- a/__tests__/unit/client.test.ts
+++ b/__tests__/unit/client.test.ts
@@ -1,0 +1,17 @@
+import { validateSelectQuery } from '../../src/client';
+
+describe('validateSelectQuery', () => {
+  it('throws on a query containing FORMAT clause', () => {
+    expect(
+() => validateSelectQuery('select * from table format JSON')).
+toThrowErrorMatchingInlineSnapshot(`"Specifying format is not supported, use \\"format\\" parameter instead."`);
+  });
+
+  it('does not throw on a "format" column name', () => {
+    expect(() => validateSelectQuery('select format from my_table')).not.toThrow();
+  });
+
+  it('does not throw on a query without "format" word', () => {
+    expect(() => validateSelectQuery('select * from my_format limit 1')).not.toThrow();
+  });
+});


jest.config.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  preset: 'ts-jest',
+  testEnvironment: 'node',
```
mshustov, line 3: I will investigate how to run the package in browser env later.

__tests__/integration/insert.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    });
+
+    const result = await Rows.json<ResponseJSON>();
+    expect(result).toEqual([
+      { id: '42', name: "hello", sku: [0, 1] },
+      { id: '43', name: "world", sku: [3, 4] },
+    ]);
+  });
+
+  it('can insert strings with non-latin symols', async() => {
```
mshustov: TODO: typo

__tests__/integration/command.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      expect(e.type).toBe('TABLE_ALREADY_EXISTS');
+      // TODO remove whitespace from end
+      expect(e.message).toBe('Table default.example already exists. ');
+    }
+  });
+
+  it.skip('can specify a parametrized query', async () => {
+    client = createClient();
+    await client.command({
+      query: 'CREATE TABLE {table_name: String} (id UInt64, name String, sku Array(UInt8), timestamp DateTime) Engine = Memory',
```
mshustov, line 51: Parameters don't seem to be supported in the `create table` query. I will investigate later.

__tests__/integration/error_parsing.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  it('returns "unknown table" error', async() => {
+    expect.assertions(5);
+
+    client = createClient();
+    try {
+      await client.select({
+        query: 'SELECT * FROM unknown_table'
+      });
+    } catch (e: any) {
+      expect(e.message).toEqual(expect.any(String));
```
mshustov, line 34: Not sure we need to assert an error message text since CH can change it without any notice.

__tests__/integration/insert.test.ts
```diff
@@ -0,0 +1,124 @@
+import { createClient, type ClickHouseClient } from '../../src';
+import Stream from 'stream';
+import type { ResponseJSON } from '../../src/clickhouse_types';
+
+describe('insert', () => {
```
mshustov, line 5: I will extend the test coverage later

src/data_formatter/formatter.ts
```diff
@@ -0,0 +1,73 @@
...
+ * Decodes a string in a ClickHouse format into a plain JavaScript object or an array of objects.
+ * @param text a string in a ClickHouse data format
+ * @param format One of the supported formats: https://clickhouse.com/docs/en/interfaces/formats/
+ */
+export function decode(text: string, format: DataFormat): any {
+  if(format === 'JSON') {
+    return JSON.parse(text)
+  } if (isStreamableJSONFamily(format)) {
+    return text.split('\n').filter(Boolean).map(l => decode(l, 'JSON'));
+  } if (format === 'CSV' || format === 'TabSeparated') {
```
mshustov, line 56: I don't think it makes sense to add parsing for these formats. The same can be achieved via `JSON*` family formats.

src/client.ts
```diff
@@ -0,0 +1,197 @@
...
+      application: config.password ?? 'clickhouse-js',
+      database: config.password ?? 'default',
+      clickhouse_settings: config.clickhouse_settings ?? {},
+      log: {
+        enable: loggingEnabled
+      }
+    };
+
+    validateConfig(this.config);
+    this.connection = createConnection(this.config);
```
mshustov, line 91: TODO: provide logger to the connection factory

__tests__/integration/insert.test.ts
```diff
@@ -0,0 +1,124 @@
...
+  let client: ClickHouseClient;
+  beforeEach(async () => {
+    client = createClient();
+    const ddl = 'CREATE TABLE test_table (id UInt64, name String, sku Array(UInt8)) Engine = Memory';
```
genzgd, line 9: Memory engines for tests might be problematic for testing against the cloud.  I'm not sure what the long term support plan for MEMORY engine tables is.  In particular because of load balancing you might not end up on the same node as you created the MEMORY table.

__tests__/integration/command.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      await client.command({ query: ddl });
+    } catch (e: any) {
+      expect(e.code).toBe('57');
+      expect(e.type).toBe('TABLE_ALREADY_EXISTS');
+      // TODO remove whitespace from end
+      expect(e.message).toBe('Table default.example already exists. ');
+    }
+  });
+
+  it.skip('can specify a parametrized query', async () => {
```
genzgd: typo -- parameterized

__tests__/integration/select.test.ts
```diff
@@ -0,0 +1,401 @@
...
+      clickhouse_settings: {
+        limit: 2
+      }
+    });
+
+    const response = await rows.text();
+    expect(response).toBe('0\n1\n');
+  });
+
+  it('can specify a parametrized query', async () => {
```
genzgd: parameterized again

__tests__/integration/select.test.ts
```diff
@@ -0,0 +1,401 @@
...
+        expect(response.statistics).toEqual(
+          expect.objectContaining({
+            elapsed: expect.any(Number),
+            rows_read: expect.any(Number),
+            bytes_read: expect.any(Number),
+          })
+        );
+      });
+
+      it.skip('returns queryId in response', async() => {
```
genzgd, line 178: Why is this skipped?

__tests__/integration/select.test.ts
```diff
@@ -0,0 +1,401 @@
...
+
+        const response = await rows.json<ResponseJSON<{number: string}>>();
+
+        expect(response.query_id).toEqual(expect.any(Number));
+      });
+    });
+  });
+
+  describe('select result asStream()', () => {
+    it('trows an exception if format is not stream-able', async () => {
```
genzgd: typo -- trows

src/client.ts
```diff
@@ -0,0 +1,197 @@
...
+    return await this.connection.ping();
+  }
+
+  async close(): Promise<void> {
+    return await this.connection.close();
+  }
+}
+
+const formatRe = /\bformat\b/i;
+function validateSelectQuery(query: string): void {
```
genzgd: unfortunately `format` is a valid column name and this would break those queries?

mshustov: That's a good point. According to  https://clickhouse.com/docs/en/sql-reference/statements/select/#syntax `FORMAT {NAME}` clause can be at the very end of the statement. I will update the regexp to check the end of the string.

__tests__/integration/command.test.ts
mshustov: I need to enable grammar check in IDE 😅

__tests__/integration/insert.test.ts
mshustov: I'm not sure which engines will be available on Cloud. Let's keep the `Memory`, we can change it later. I added a comment to the Cloud testing issue to consider another table engine. https://github.com/ClickHouse/clickhouse-js/issues/3

__tests__/integration/select.test.ts
mshustov: not implemented yet. I removed the logic for now

PR #19, opened 2022-08-05
use URL instead of a string in "ConnectionParams"

mshustov:
creates URL instance during config validation and passes it down to the HTTP adapter

### DIFF
--- a/__tests__/unit/client.test.ts
+++ b/__tests__/unit/client.test.ts
@@ -3,7 +3,9 @@ import { validateSelectQuery, createClient } from '../../src/client';
 
 describe('createClient', () => {
   it('throws on incorrect "host" config value', () => {
-    expect(() => createClient({ host: 'foo' })).to.throw(/Invalid URL/);
+    expect(() => createClient({ host: 'foo' })).to.throw(
+      'Configuration parameter "host" contains malformed url.'
+    );
   });
 });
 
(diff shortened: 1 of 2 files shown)


__tests__/unit/client.test.ts
```diff
@@ -1,5 +1,11 @@
 import { expect } from 'chai';
-import { validateSelectQuery } from '../../src/client';
+import { validateSelectQuery, createClient } from '../../src/client';
+
+describe('createClient', () => {
+  it('throws on incorrect "host" config value', () => {
+    expect(() => createClient({ host: 'foo' })).to.throw(/Invalid URL/);
```
mshustov: Ok, we can improve it by providing a detailed message that a value in `host` parameter is invalid

__tests__/unit/transform_url.test.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  it('does not mutate an original url', () => {
+    const url = new URL('http://clickhouse.com');
+    const newUrl = transformUrl({
+      url,
+      pathname: 'foo',
+    });
+    expect(newUrl.toString()).to.equal('http://clickhouse.com/foo');
+    expect(url.toString()).to.equal('http://clickhouse.com/');
+  });
```
slvrtrn, line 32: 👍

Issue #3, opened 2022-07-26
Automate testing against Cloud


mshustov:
Run integration tests against ClickHouse Cloud `dev` env.
It should be possible to run tests against any provided URL. 
We might need to change the table engine in tests from `Memory` to one supported by Cloud.

PR #22, opened 2022-08-05
Allow tests to use ClickHouse Cloud, add simple HTTPS support

slvrtrn:
Implements https://github.com/ClickHouse/clickhouse-js/issues/6, partially https://github.com/ClickHouse/clickhouse-js/issues/3

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -17,6 +17,7 @@ jobs:
       fail-fast: true
       matrix:
         node: [14, 16]
+        clickhouse: [head, latest]
 
     services:
       clickhouse:


.github/workflows/run-tests.yml
```diff
@@ -16,11 +16,8 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        node-version:
-          - 14.x
-          - 16.x
+        node: [ 14 ]
         clickhouse:
-          - head
           - latest
```
slvrtrn: Need to investigate why we have failures using Cloud env with 4 parallel test runs when using a proper matrix

mshustov: Let's create a dedicated issue to fix the problem in a follow-up PR

__tests__/utils/client.ts
```diff
@@ -0,0 +1,61 @@
+import {
+  ClickHouseClient,
+  ClickHouseClientConfigOptions,
+  createClient,
+} from '../../src';
+import { randomUUID } from 'crypto';
+
+export type TestClientConfiguration = ClickHouseClientConfigOptions & {
+  useCloud?: boolean;
```
mshustov: Let's add a comment it's a temporary flag until we make all the tests compatible with Cloud.

PR #31, opened 2022-08-08
Fix parse_error to handle messages with parenthesis 

slvrtrn:


src/error/parse_error.ts
```diff
@@ -1,5 +1,5 @@
 const errorRe =
-  /(Code|Error): (?<code>\d+).*Exception: (?<message>.+?)\((?<type>\w+?)\)/is;
+  /(Code|Error): (?<code>\d+).*Exception: (?<message>.+? ?)\((?<type>(?=[A-Z]{3})[A-Z0-9_]+?)\)/is;
```
mshustov: Could you add a test case with an error message containing numbers? `S3_ERROR`, for example

PR #33, opened 2022-08-09
Local ClickHouse cluster configuration

slvrtrn:
On top of https://github.com/ClickHouse/clickhouse-js/pull/22, to keep it cleaner for the review.
Now we can run a ClickHouse cluster locally, with 2 nodes and Clickhouse Keeper (no Zookeeper)

Nginx serves as a load balancer to expose a single ClickHouse endpoint like it is done in the Cloud.

### DIFF
--- a/__tests__/utils/client.ts
+++ b/__tests__/utils/client.ts
@@ -7,9 +7,9 @@ import {
 import { guid } from './guid';
 
 export enum TestEnv {
-  Cloud = 'CLOUD',
-  LocalSingleNode = 'LOCAL_SINGLE_NODE',
-  LocalCluster = 'LOCAL_CLUSTER',
+  Cloud = 'cloud',
+  LocalSingleNode = 'local_single_node',
+  LocalCluster = 'local_cluster',
 }
 
 export function createTestClient(


docker-compose.cluster.yml
```diff
@@ -0,0 +1,46 @@
...
+        soft: 262144
+        hard: 262144
+    hostname: clickhouse1
+    container_name: clickhouse-js-clickhouse-server-node-1
+    ports:
+      - '8124:8123'
+      - '9000:9000'
+      - '9181:9181'
+    volumes:
+      - './.docker/clickhouse/cluster/config1.xml:/etc/clickhouse-server/config.xml'
```
mshustov: Optional: what if we give them more descriptive names? `config1.xml` --> `server_1_config.xml`?

__tests__/integration/insert.test.ts
```diff
@@ -2,20 +2,43 @@ import { expect } from 'chai';
...
+import { TestEnv } from '../utils/client';
 
 describe('insert', () => {
   let client: ClickHouseClient;
   let tableName: string;
   beforeEach(async () => {
-    client = await createTestClient();
+    client = await createTestClient({
+      clickhouse_settings: {
+        insert_quorum: 2,
```
mshustov: Is it still necessary for the tests?

src/client.ts
```diff
@@ -140,13 +141,16 @@ export class ClickHouseClient {
...
+    const format = params.format ?? 'JSON';
+    const query = formatCommandQuery(params.query, format);
 
-    await this.connection.command({
+    const stream = await this.connection.command({
       query,
       ...this.getBaseParams(params),
     });
+
+    return new Rows(stream, format);
```
mshustov, line 153: Thinking loudly: we need to rename the result of the operation. not a blocker for the current PR

__tests__/utils/client.ts
```diff
@@ -57,6 +68,19 @@ function getFromEnv(key: string): string {
...
 
-function isClickHouseCloudEnabled() {
-  return process.env['CLICKHOUSE_CLOUD_ENABLED'] === 'true';
+export function getClickHouseTestEnvironment(): TestEnv {
+  let env;
+  switch (process.env['CLICKHOUSE_TEST_ENVIRONMENT']) {
+    case 'CLOUD':
+      env = TestEnv.Cloud;
+      break;
+    case 'LOCAL_CLUSTER':
```
mshustov, line 77: Where do we set it? Shouldn't we add a command to `package.json` to run the tests against a local cluster? Can be done in follow-up, though.

__tests__/integration/insert.test.ts
slvrtrn: removed

__tests__/utils/client.ts
slvrtrn: currently in GHA only: https://github.com/ClickHouse/clickhouse-js/blob/97db905e487be69ecc472d91a4b2230fe50654cb/.github/workflows/run-tests.yml#L55

but I agree

PR #40, opened 2022-08-11
Use Jest instead of Mocha

### LINKED ISSUE #30: Use Jest instead of Mocha
Related to https://github.com/ClickHouse/clickhouse-js/issues/17

- use better promise matchers
- verify that the tests work in the browser environment

slvrtrn:
Resolves #30 

- [x] Jest (everything is working)
- [x] Use Jest mocks, remove Sinon
- [x] Remove the rest of `done` callbacks
- [x] Re-introduce random database creation (need to figure out the best way of doing the global setup for integration tests only)
- [x] Review test patterns matching, as those are regex now (`npm test` does not work as expected, other test commands are fine)

BTW, current coverage (integration tests):

```
=============================== Coverage summary ===============================
Statements   : 87.22% ( 280/321 )
Branches     : 76.16% ( 131/172 )
Functions    : 88.76% ( 79/89 )
Lines        : 90% ( 270/300 )
================================================================================
```

### DIFF
--- a/__tests__/utils/client.ts
+++ b/__tests__/utils/client.ts
@@ -54,9 +54,13 @@ export async function createRandomDatabase(
   client: ClickHouseClient
 ): Promise<string> {
   const databaseName = `clickhousejs__${guid()}`;
+  let maybeOnCluster = '';
+  if (getClickHouseTestEnvironment() === TestEnv.LocalCluster) {
+    maybeOnCluster = `ON CLUSTER '{cluster}'`;
+  }
   await (
     await client.command({
-      query: `CREATE DATABASE IF NOT EXISTS ${databaseName}`,
+      query: `CREATE DATABASE IF NOT EXISTS ${databaseName} ${maybeOnCluster}`,
     })
   ).text();
   console.log(`Created database ${databaseName}`);


__tests__/unit/http_adapter.ts
```diff
@@ -78,23 +21,17 @@ describe('HttpAdapter', () => {
...
-          query: 'SELECT * from system.numbers LIMIT 5',
+        const selectPromise = adapter.select({
+          query: 'SELECT * FROM system.numbers LIMIT 5',
         });
-        expect(httpRequestStub?.callCount).of.equal(1);
 
-        const calledWith: { headers: Record<string, string> } =
-          httpRequestStub?.getCall(0).args[0];
+        const responseBody = 'foobar';
+        await emitCompressedBody(request, responseBody);
```
slvrtrn: @mshustov please confirm that I got the logic right in these updated tests; there were timeouts in the first three until I emitted some data.

__tests__/utils/guid.ts
```diff
@@ -1,5 +1,5 @@
-import { randomUUID } from 'crypto';
+import { v4 as uuid_v4 } from 'uuid';
 
 export function guid() {
-  return randomUUID().replace(/-/g, '');
```
slvrtrn, line 4: Jest did not like the default `crypto` module at all. Added `uuid` as a dev dependency.

package.json
```diff
@@ -7,11 +7,11 @@
     "build": "rm -rf dist; webpack --mode=development",
     "compile": "tsc --project tsconfig.eslint.json --noEmit",
     "lint": "eslint . --ext .ts",
-    "test": "mocha __tests__/**/*.test.ts --exit",
-    "test:unit": "mocha __tests__/unit/**/*.test.ts --exit",
-    "test:integration": "mocha __tests__/integration/**/*.test.ts --exit --require ts-node/register --require __tests__/utils/global_setup.ts",
-    "test:integration:local_cluster": "CLICKHOUSE_TEST_ENVIRONMENT=local_cluster mocha __tests__/integration/**/*.test.ts --exit --require ts-node/register --require __tests__/utils/global_setup.ts",
-    "test:integration:cloud": "CLICKHOUSE_TEST_ENVIRONMENT=cloud mocha __tests__/integration/**/*.test.ts --exit --require ts-node/register --require __tests__/utils/global_setup.ts",
+    "test": "jest --testPathPattern=__tests__",
```
slvrtrn: this is a bit broken - it tries to run `utils` module files, which contain no tests.

tsconfig.eslint.json
```diff
@@ -1,4 +1,8 @@
 {
   "extends": "./tsconfig.json",
-  "include": ["__tests__/**/*.ts"]
+  "include": ["__tests__/**/*.ts"],
+  "compilerOptions": {
+    "noUnusedLocals": false,
+    "noUnusedParameters": false,
```
slvrtrn, line 6: disabled it as a compiler option for tests (ESlint will catch it anyway) cause it is super annoying during debugging of the tests to comment out imports or unused variables all the time

__tests__/integration/abort_request.test.ts
```diff
@@ -53,7 +32,7 @@ describe('abort request', () => {
           abort_signal: controller.signal as AbortSignal,
         })
         .catch((error: Error) => {
-          expect(error.message).to.equal('The request was aborted.');
+          expect(error.message).toBe('The request was aborted.');
```
slvrtrn: that's how it was translated by https://github.com/skovhus/jest-codemods#usage-cli

mshustov: That's the correct refactoring

.eslintignore
```diff
@@ -0,0 +1 @@
+*.js
```
mshustov: Why is it necessary? We can have some scripts in the plain js. Eslint should work for them too.

__tests__/integration/config.test.ts
```diff
@@ -1,31 +1,32 @@
...
 import { createTestClient } from '../utils';
 
 describe('config', () => {
-  before(function () {
-    if (process.env.browser) {
-      this.skip();
-    }
+  beforeAll(function () {
+    // FIXME: Jest does not seem to have it
+    // if (process.env.browser) {
```
mshustov, line 7: It doesn't. I set it manually https://github.com/ClickHouse/clickhouse-js/blob/main/webpack.config.js#L49

__tests__/unit/http_adapter.ts
mshustov: LGTM. Optional: we have a dedicated test for the body compression, I'd consider removing `expect(await getAsText(await selectPromise)).toBe(responseBody);` assertion to keep the test focused

.eslintignore
slvrtrn: Ah, that was because ESLint was complaining about module resolution in the JS files in the root. I guess it's a leftover. Will fix.

PR #44, opened 2022-08-12
Add `toSearchParams` unit test

slvrtrn:
A leftover from #40

### DIFF
--- a/src/connection/adapter/http_search_params.ts
+++ b/src/connection/adapter/http_search_params.ts
@@ -1,7 +1,7 @@
 import { type ClickHouseSettings } from '../../clickhouse_types';
 import { formatQueryParams, formatQuerySettings } from '../../data_formatter/';
 
-export type ToSearchParamsOptions = {
+type ToSearchParamsOptions = {
   database: string;
   clickhouse_settings?: ClickHouseSettings;
   query_params?: Record<string, unknown>;


src/connection/adapter/http_search_params.ts
```diff
@@ -1,18 +1,26 @@
 import { type ClickHouseSettings } from '../../clickhouse_types';
 import { formatQueryParams, formatQuerySettings } from '../../data_formatter/';
 
+export type ToSearchParamsOptions = {
```
mshustov: optional: no need to export an internal type

src/connection/adapter/http_search_params.ts
```diff
@@ -1,18 +1,26 @@
...
+  query_params,
+  clickhouse_settings,
+}: ToSearchParamsOptions): URLSearchParams | undefined {
   if (
-    settings === undefined &&
+    clickhouse_settings === undefined &&
     query_params === undefined &&
-    query === undefined
+    query === undefined &&
+    database === 'default'
```
mshustov, line 23: Optional nit: The default database value leaks from the config. Maybe it's OK always to provide `database` via query explicitly.

slvrtrn: Hmm, that's a tricky one then. Let's come back to this after the first app usage attempt.

Issue #28, opened 2022-08-07
Extend the list of supported formats 


mshustov:
Add `JSON*`, `CSV*`, `TabSeparated*` formats that can be supported by the client in the current implementation https://github.com/ClickHouse/clickhouse-js/blob/6caae6067928700847a524793d8db535cd631976/src/data_formatter/formatter.ts#L2-L22
https://clickhouse.com/docs/en/interfaces/formats/

- [x] JSON
- [x] CSV
- [x] TSV
- [x] Custom separated

slvrtrn:
JSON formats for streaming are fully covered in #83

PR #47, opened 2022-08-13
Experimental schema support / type inference

slvrtrn:
An implementation of the ideas from https://github.com/ClickHouse/integrations/issues/21

### DIFF
--- a/src/schema/types.ts
+++ b/src/schema/types.ts
@@ -41,7 +41,7 @@ export type Type =
   | IPv4
   | IPv6
 
-export type UInt8 = {
+export interface UInt8 {
   underlying: number
   type: 'UInt8'
 }
@@ -51,7 +51,7 @@ export const UInt8 = {
     return 'UInt8'
   },
 } as UInt8
-export type UInt16 = {
+export interface UInt16 {
   type: 'UInt16'
   underlying: number
 }
@@ -61,7 +61,7 @@ export const UInt16 = {
     return 'UInt16'
   },
 } as UInt16
-export type UInt32 = {
+export interface UInt32 {
   type: 'UInt32'
   underlying: number
 }
@@ -71,7 +71,7 @@ export const UInt32 = {
     return 'UInt32'
   },
 } as UInt32
-export type UInt64 = {
+export interface UInt64 {
   underlying: string
   type: 'UInt64'
 }
@@ -91,7 +91,7 @@ export const UInt64 = {
     return 'UInt64'
   },
 } as UInt64
-export type UInt128 = {
+export interface UInt128 {
   type: 'UInt128'
   underlying: string
 }
@@ -105,7 +105,7 @@ export const UInt128 = {
     return 'UInt128'
   },
 } as UInt128
-export type UInt256 = {
+export interface UInt256 {
   type: 'UInt256'
   underlying: string
 }
@@ -120,7 +120,7 @@ export const UInt256 = {
   },
 } as UInt256
 
-export type Int8 = {
+export interface Int8 {
   underlying: number
   type: 'Int8'
 }
@@ -130,7 +130,7 @@ export const Int8 = {
     return 'Int8'
   },
 } as Int8
-export type Int16 = {
+export interface Int16 {
   type: 'Int16'
   underlying: number
 }
@@ -140,7 +140,7 @@ export const Int16 = {
     return 'Int16'
   },
 } as Int16
-export type Int32 = {
+export interface Int32 {
   type: 'Int32'
   underlying: number
 }
@@ -151,7 +151,7 @@ export const Int32 = {
   },
 } as Int32
 
-export type Int64 = {
+export interface Int64 {
   underlying: string
   type: 'Int64'
 }
@@ -171,7 +171,7 @@ export const Int64 = {
     return 'Int64'
   },
 } as Int64
-export type Int128 = {
+export interface Int128 {
   type: 'Int128'
   underlying: string
 }
@@ -185,7 +185,7 @@ export const Int128 = {
     return 'Int128'
   },
 } as Int128
-export type Int256 = {
+export interface Int256 {
   type: 'Int256'
   underlying: string
 }
@@ -200,7 +200,7 @@ export const Int256 = {
   },
 } as Int256
 
-export type Float32 = {
+export interface Float32 {
   type: 'Float32'
   underlying: number
 }
@@ -210,7 +210,7 @@ export const Float32 = {
     return 'Float32'
   },
 } as Float32
-export type Float64 = {
+export interface Float64 {
   type: 'Float64'
   underlying: number
 }
@@ -221,7 +221,7 @@ export const Float64 = {
   },
 } as Float64
 
-export type Decimal = {
+export interface Decimal {
   type: 'Decimal'
   underlying: number
 }
@@ -261,7 +261,7 @@ export const Decimal = ({
     },
   } as Decimal)
 
-export type Bool = {
+export interface Bool {
   type: 'Bool'
   underlying: boolean
 }
@@ -272,7 +272,7 @@ export const Bool = {
   },
 } as Bool
 
-export type String = {
+export interface String {
   type: 'String'
   underlying: string
 }
@@ -283,7 +283,7 @@ export const String = {
   },
 } as String
 
-export type FixedString = {
+export interface FixedString {
   type: 'FixedString'
   underlying: string
 }
@@ -295,7 +295,7 @@ export const FixedString = (bytes: number) =>
     },
   } as FixedString)
 
-export type UUID = {
+export interface UUID {
   type: 'UUID'
   underlying: string
 }
@@ -311,7 +311,7 @@ type StandardEnum<T> = {
   [n: number]: string
 }
 
-export type Enum<T extends StandardEnum<unknown>> = {
+export interface Enum<T extends StandardEnum<unknown>> {
   type: 'Enum'
   underlying: keyof T
 }
@@ -336,7 +336,7 @@ type LowCardinalityDataType =
   | Float
   | Date
   | DateTime
-export type LowCardinality<T extends LowCardinalityDataType> = {
+export interface LowCardinality<T extends LowCardinalityDataType> {
   type: 'LowCardinality'
   underlying: T['underlying']
 }
@@ -348,7 +348,7 @@ export const LowCardinality = <T extends LowCardinalityDataType>(type: T) =>
     },
   } as LowCardinality<T>)
 
-export type Array<T extends Type> = {
+export interface Array<T extends Type> {
   type: 'Array'
   underlying: globalThis.Array<T['underlying']>
 }
@@ -375,7 +375,7 @@ type NullableType =
   | Date32
   | IPv4
   | IPv6
-export type Nullable<T extends NullableType> = {
+export interface Nullable<T extends NullableType> {
   type: 'Nullable'
   underlying: T['underlying'] | null
 }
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -50,6 +50,7 @@ export interface CommandParams extends BaseParams {
 export interface InsertParams extends BaseParams {
   table: string
   values: ReadonlyArray<any> | Stream.Readable
+  format?: DataFormat
```
mshustov, line 56: Are you going to support other formats from `DataFormat` besides `JSONCompactEachRow` and `JSONEachRow`? If not, let's adjust typings. Right now, `insert` throws on other formats
https://github.com/ClickHouse/clickhouse-js/blob/4ed7e5f27e0d3f123836daa978fbb1292d1dd603/src/data_formatter/formatter.ts#L83

src/client.ts
```diff
@@ -220,10 +222,10 @@ function encodeValues(
   if (isStream(values)) {
     return Stream.pipeline(
       values,
-      mapStream(function (value: any) {
+      mapStream((value: unknown) => {
         return encode(value, format)
       }),
-      function pipelineCb(err) {
+      (err) => {
```
mshustov: JFYI: I used `function pipelineCb` to keep the function name in the error stack trace

src/schema/schema.ts
```diff
@@ -0,0 +1,13 @@
+import { Shape } from './common'
+
+export class Schema<S extends Shape> {
+  constructor(public readonly shape: S) {}
+
+  toString(): string
+  toString(delimiter?: string): string
```
mshustov: nit: can be removed

__tests__/integration/schema.test.ts
```diff
@@ -0,0 +1,435 @@
...
+          ['d1']
+        )
+      })
+
+      it('should insert and select it back', async () => {
+        await assertInsertAndSelect(table, value)
+      })
+    })
+
+    // describe('Decimal', () => {
```
mshustov: What if we keep it uncommented but skipped for demo purposes?

src/schema/types.ts
```diff
@@ -0,0 +1,462 @@
...
+  AggregateFunction
+  SimpleAggregateFunction
+  Nested
+  Special Data Types
+  Geo (experimental)
+  Multi-word Types
+  Better Date(Time) parsing/handling, including timezones
+  Tuple
+  Named tuple
+  Decimal (without precision loss) - see https://github.com/ClickHouse/ClickHouse/issues/21875
```
mshustov: Could you add an extended comment on why it's not supported currently?

src/schema/types.ts
```diff
@@ -0,0 +1,462 @@
...
+  | Enum<any>
+  | LowCardinality<any>
+  | Date
+  | Date32
+  | DateTime
+  | DateTime64
+  | IPv4
+  | IPv6
+
+export type UInt8 = {
```
mshustov: Is there a reason to prefer `type` over `interface`? https://www.typescriptlang.org/play#example/types-vs-interfaces

I also see you disabled `@typescript-eslint/ban-types` rule.

src/schema/types.ts
```diff
@@ -0,0 +1,462 @@
...
+  type: 'UInt32',
+  toString(): string {
+    return 'UInt32'
+  },
+} as UInt32
+
+export type UInt64 = {
+  // Max UInt64:               18446744073709551615
+  // Number.MAX_SAFE_INTEGER:  9007199254740991
+  // can be cast to number by disabling `output_format_json_quote_64bit_integers` CH setting
```
mshustov: Could you add the comment in the form of TSDoc to provide better suggestions in IDE? as proposed here https://github.com/ClickHouse/clickhouse-js/pull/51#discussion_r948809969

src/schema/types.ts
```diff
@@ -0,0 +1,462 @@
...
+  ({
+    type: 'Map',
+    toString(): string {
+      return `Map(${k.toString()}, ${v.toString()})`
+    },
+  } as Map<K, V>)
+
+export type Date = {
+  type: 'Date'
+  underlying: string // '1970-01-01' to '2149-06-06'
```
mshustov, line 414: Let move these remarks under TSDoc as well

src/schema/where.ts
```diff
@@ -0,0 +1,48 @@
...
+  return {
+    toString(): string {
+      return `(${expr.join(' OR ')})`
+    },
+    type: 'Or',
+  }
+}
+
+function formatValue(value: any): string {
+  if (value === null || value === undefined) {
```
mshustov, line 42: Why should we handle `undefined` as `NULL`?

src/schema/where.ts
```diff
@@ -0,0 +1,48 @@
+import { Shape } from './common'
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export interface WhereExpr<S extends Shape> {
```
mshustov, line 4: Are there any tests for `where`?

src/schema/schema.ts
```diff
@@ -0,0 +1,13 @@
...
+  toString(delimiter?: string): string {
+    return Object.entries(this.shape)
+      .map(([column, type]) => `${column} ${type.toString()}`)
```
mshustov, line 8: We might need to escape name for columns with ` for cases when they contain special characters or a reserved word
```sql
CREATE TABLE baz (`foo.bar` UInt64) engine=Memory
CREATE TABLE bar (`key` UInt64) engine=Memory
```

examples/schema/simple_schema.ts
```diff
@@ -0,0 +1,48 @@
+import * as ch from '../../src/schema'
+import { Infer, InsertStream } from '../../src/schema'
+import { createClient } from '../../src'
+
+export default async () => {
```
mshustov: nit: it's not used outside. let's remove `export`. You can just declare a function and call it in the example file

src/schema/compact.ts
```diff
@@ -0,0 +1,21 @@
...
+  const compacted = []
+  for (const key in shape) {
+    if (key in value) {
+      compacted.push(value[key])
+    }
+  }
+  return compacted
+}
+
+export function decompactJson<S extends Shape>(shape: S, [row]: [unknown[]]) {
```
mshustov, line 13: It would be interesting to measure the impact of these functions on the runtime performance. Not now, we don't infrastructure, but we need to keep in mind the overhead of copying a potentially large stream of json objects.

src/schema/engines.ts
```diff
@@ -0,0 +1,117 @@
+// See https://clickhouse.com/docs/en/engines/table-engines/
+
+interface Engine {
```
mshustov: What is the purpose of this type?

__tests__/integration/schema.test.ts
```diff
@@ -0,0 +1,435 @@
...
+        values,
+      })
+      const result = await (await table.select()).json()
+      expect(result).toEqual(values)
+    })
+
+    it('should insert and select data using streams', async () => {
+      const values = new ch.InsertStream<Value>()
+      values.add(value1)
+      values.add(value2)
```
mshustov: We can test that TS complains about a wrong type:
```typescript
const values = new ch.InsertStream<Value>()
// @ts-expect-error
values.add([null]);
values.complete();
```

src/schema/table.ts
```diff
@@ -0,0 +1,112 @@
...
+    const query = QueryFormatter.select(this.options, where, columns, order_by)
+    const rows = await this.client.command({
+      query,
+      clickhouse_settings,
+      abort_signal,
+      format: 'JSONEachRow',
+    })
+
+    const stream = rows.asStream()
+    async function* asyncGenerator() {
```
mshustov, line 91: Maybe we need to declare `Iterator` explicitly? https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-6.html#stricter-generators

src/schema/table.ts
```diff
@@ -0,0 +1,112 @@
...
+      for await (const row of stream) {
+        const value = (row as Row).json() as unknown[]
+        yield value[0] as Infer<S> // FIXME why we have an array here?
+      }
+    }
+
+    return {
+      asyncGenerator,
```
mshustov, line 99: why not [Sumbol.AsyncItetator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncIterator)?

__tests__/integration/schema.test.ts
```diff
@@ -0,0 +1,435 @@
...
+      setTimeout(() => values.complete(), 100)
+
+      await table.insert({
+        values,
+      })
+
+      const result: Value[] = []
+      const { asyncGenerator } = await table.select()
```
mshustov: Could you add a test that errors bubble up to the `asyncGenerator` consumer?

__tests__/integration/schema.test.ts
```diff
@@ -0,0 +1,435 @@
...
+        n2: '2022-04-30',
+      }
+
+      let table: ch.Table<typeof shape>
+      beforeEach(async () => {
+        table = await createTableWithSchema(
+          client,
+          new ch.Schema(shape),
+          tableName,
+          ['id']
```
mshustov: So, we can't use `shape.id` here?

__tests__/integration/schema.test.ts
```diff
@@ -0,0 +1,435 @@
...
+
+      const shape = {
+        id: ch.Int32, // to preserve the order of values
+        e: ch.Enum(MyEnum),
+      }
+      const values: ch.Infer<typeof shape>[] = [
+        { id: 1, e: MyEnum.Bar },
+        { id: 2, e: MyEnum.Qux },
+        { id: 3, e: MyEnum.Foo },
+        { id: 4, e: MyEnum.Qaz },
```
mshustov: a test with a unassignable value would be 🔥

__tests__/unit/query_formatter.ts
```diff
@@ -0,0 +1,56 @@
...
+    const tableOptions = {
+      name: 'my_table',
+      schema,
+    }
+    expect(
+      QueryFormatter.createTable(tableOptions, {
+        engine: ch.MergeTree(),
+        if_not_exists: true,
+        on_cluster: '{cluster}',
+        order_by: ['foo', 'bar'],
```
mshustov, line 38: Should we validate that it throws on invalid values like `[]`? can be done in a follow-up

__tests__/unit/query_formatter.ts
```diff
@@ -0,0 +1,56 @@
+import * as ch from '../../src/schema'
+import { QueryFormatter } from '../../src/schema/query_formatter'
+
+describe('QueryFormatter', () => {
+  it('should render a simple CREATE TABLE statement', async () => {
```
mshustov, line 5: Could you add tests for `select` statement?

src/schema/types.ts
slvrtrn: `ban-types` reacts on the collision with the built-in types - String, Map etc, that's why it is disabled
No strong preference for `type` vs `interface`

src/schema/where.ts
slvrtrn: not yet. Will add.

src/client.ts
slvrtrn: let's do this cleanup as a follow-up PR that will cover #28

__tests__/integration/schema.test.ts
slvrtrn: we could try, but types are inferred properly anyway and you have full IDE support

image

__tests__/integration/schema.test.ts
slvrtrn: Unfortunately, we will have to export Decimal as a CH type then. I'd prefer not to since my current implementation is not stable.

PR #49, opened 2022-08-17
Integration tests with a read only user

slvrtrn:
As expected: we cannot use response http compression for a read-only user

### DIFF
--- a/__tests__/integration/read_only_user.test.ts
+++ b/__tests__/integration/read_only_user.test.ts
@@ -10,7 +10,7 @@ describe('read only user', () => {
     const database = getTestDatabaseName()
     const defaultClient = createTestClient()
 
-    const { username } = await createReadOnlyUser(defaultClient)
+    const { username, password } = await createReadOnlyUser(defaultClient)
 
     // Populate some test table to select from
     tableName = `read_only_user_data_${guid()}`
@@ -25,6 +25,7 @@ describe('read only user', () => {
     // Create a client that connects read only user to the test database
     client = createTestClient({
       username,
+      password,
       database,
       clickhouse_settings: {
         insert_quorum: undefined,
--- a/__tests__/integration/fixtures/read_only_user.ts
+++ b/__tests__/integration/fixtures/read_only_user.ts
@@ -8,6 +8,7 @@ import { ClickHouseClient } from '../../../src'
 
 export async function createReadOnlyUser(client: ClickHouseClient) {
   const username = `clickhousejs__read_only_user_${guid()}`
+  const password = guid()
   const database = getTestDatabaseName()
   const env = getClickHouseTestEnvironment()
   let createUser: string
@@ -17,6 +18,7 @@ export async function createReadOnlyUser(client: ClickHouseClient) {
     case TestEnv.LocalSingleNode:
       createUser = `
           CREATE USER ${username} 
+          IDENTIFIED WITH sha256_password BY '${password}'
           DEFAULT DATABASE ${database}
           SETTINGS readonly = 1
         `
@@ -30,6 +32,7 @@ export async function createReadOnlyUser(client: ClickHouseClient) {
       createUser = `
           CREATE USER ${username} 
           ON CLUSTER '{cluster}'
+          IDENTIFIED WITH sha256_password BY '${password}'
           DEFAULT DATABASE ${database}
           SETTINGS readonly = 1
         `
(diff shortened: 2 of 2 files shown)


src/client.ts
```diff
@@ -185,10 +185,12 @@ function formatSelectQuery(query: string, format: DataFormat): string {
...
 
-// it is a duplicate of `formatSelectQuery`, but it might differ in the future
-function formatCommandQuery(query: string, format: DataFormat): string {
+function formatCommandQuery(query: string, format: DataFormat | null): string {
   query = query.trim()
-  return query + ' \nFORMAT ' + format
+  if (format !== null) {
+    return query + ' \nFORMAT ' + format
+  }
+  return query
```
slvrtrn, line 197: Some commands such as GRANT do not support the FORMAT clause, so I needed some way to disable `FORMAT ${format}` entirely. For now, it is an explicit null used as follows: 

```
await client.command({
  query,
  format: null
})
```
@mshustov wdyt?

src/client.ts
```diff
@@ -139,15 +139,15 @@ export class ClickHouseClient {
...
+    const format = params.format === undefined ? 'JSON' : params.format
     const query = formatCommandQuery(params.query, format)
 
     const stream = await this.connection.command({
       query,
       ...this.getBaseParams(params),
     })
 
-    return new Rows(stream, format)
+    return new Rows(stream, 'JSON')
```
slvrtrn: oops, will fix

src/client.ts
mshustov: Ok. Another option might be explicit `false`.

src/client.ts
```diff
@@ -44,7 +44,7 @@ export interface SelectParams extends BaseParams {
 
 export interface CommandParams extends BaseParams {
   query: string
-  format?: DataFormat
+  format?: DataFormat | null
```
mshustov: Could you add a comment explaining the purpose of `null`?

src/connection/adapter/http_search_params.ts
```diff
@@ -35,7 +35,9 @@ export function toSearchParams({
 
   if (clickhouse_settings !== undefined) {
     for (const [key, value] of Object.entries(clickhouse_settings)) {
-      params.set(key, formatQuerySettings(value))
+      if (value !== undefined) {
```
mshustov, line 38: Could you add a unit test?

__tests__/integration/read_only_user.test.ts
```diff
@@ -0,0 +1,134 @@
...
+            `To execute this query it's necessary to have grant INSERT`
+        ),
+      })
+    )
+  })
+
+  // TODO: enabled by default
+  //  we can try to restrict the access using https://clickhouse.com/docs/en/sql-reference/statements/create/row-policy/
+  //  as suggested in https://github.com/ClickHouse/ClickHouse/issues/2451#issuecomment-1162497824
+  it.skip('should fail to query system tables', async () => {
```
mshustov: What prevents us from specifying a row policy and un-skipping the test? Sorry, it's not entirely clear from the comment.

__tests__/integration/read_only_user.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    await expect(
+      client.insert({
+        table: tableName,
+        values: [[43, 'foobar', [5, 25]]],
+      })
+    ).rejects.toEqual(
+      expect.objectContaining({
+        message: expect.stringContaining(
+          'Not enough privileges. ' +
+            `To execute this query it's necessary to have grant INSERT`
```
mshustov: Optional: I'm on the fence with such assertions. The more details it is, the higher risk of flakiness due to some changes in string content. IMO a minimalistic assertion against `Not enough privileges` is okay.

__tests__/integration/read_only_user.test.ts
```diff
@@ -0,0 +1,134 @@
...
+  beforeAll(async () => {
+    username = `clickhousejs__read_only_user_${guid()}`
+    const database = getTestDatabaseName()
+    const defaultClient = createTestClient()
+    const env = getClickHouseTestEnvironment()
+
+    // Bootstrap read only user
+    let createUser: string
+    let grant: string
+    switch (env) {
```
mshustov: optional: I'd suggest extracting it in a test helper to reduce the cognitive load for a developer reading the test.

src/client.ts
slvrtrn: Yep, I like explicit `false` better

PR #51, opened 2022-08-17
Complete ClickHouse server settings typings

### LINKED ISSUE #27: Add typings for ClickHouse settings
https://clickhouse.com/docs/en/operations/settings/settings/

slvrtrn:
Resolves #27

### DIFF
--- a/__tests__/unit/to_search_params.test.ts
+++ b/__tests__/unit/to_search_params.test.ts
@@ -18,6 +18,7 @@ describe('toSearchParams', () => {
         insert_quorum: '2',
         distributed_product_mode: 'global',
         limit: '42',
+        allow_nondeterministic_mutations: undefined, // will be omitted
       },
     })!
     expect(toSortedArray(params)).toEqual([
@@ -59,6 +60,7 @@ describe('toSearchParams', () => {
       clickhouse_settings: {
         extremes: 1,
         enable_optimize_predicate_expression: 0,
+        wait_end_of_query: 1,
       },
       query_params: {
         qaz: 'qux',
@@ -72,6 +74,7 @@ describe('toSearchParams', () => {
       ['extremes', '1'],
       ['param_qaz', 'qux'],
       ['query', 'SELECT * FROM system.query_log'],
+      ['wait_end_of_query', '1'],
     ])
   })
 })
--- a/src/client.ts
+++ b/src/client.ts
@@ -43,7 +43,11 @@ export interface SelectParams extends BaseParams {
 
 export interface CommandParams extends BaseParams {
   query: string
-  format?: DataFormat
+  /**
+   * Use 'false' if your command does not support FORMAT statement,
+   * and it needs to be omitted from the query
+   */
+  format?: DataFormat | false
 }
 
 export interface InsertParams extends BaseParams {
@@ -138,7 +142,7 @@ export class ClickHouseClient {
   }
 
   async command(params: CommandParams): Promise<Rows> {
-    const format = params.format ?? 'JSON'
+    const format = params.format === undefined ? 'JSON' : params.format
     const query = formatCommandQuery(params.query, format)
 
     const stream = await this.connection.command({
@@ -146,7 +150,7 @@ export class ClickHouseClient {
       ...this.getBaseParams(params),
     })
 
-    return new Rows(stream, format)
+    return new Rows(stream, format as any)
   }
 
   async insert(params: InsertParams): Promise<void> {
@@ -184,10 +188,12 @@ function formatSelectQuery(query: string, format: DataFormat): string {
   return query + ' \nFORMAT ' + format
 }
 
-// it is a duplicate of `formatSelectQuery`, but it might differ in the future
-function formatCommandQuery(query: string, format: DataFormat): string {
+function formatCommandQuery(query: string, format: DataFormat | false): string {
   query = query.trim()
-  return query + ' \nFORMAT ' + format
+  if (format !== false) {
+    return query + ' \nFORMAT ' + format
+  }
+  return query
 }
 
 function validateInsertValues(
--- a/.docker/clickhouse/single_node/config.xml
+++ b/.docker/clickhouse/single_node/config.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0"?>
+<clickhouse>
+
+    <http_port>8123</http_port>
+
+    <users_config>users.xml</users_config>
+    <default_profile>default</default_profile>
+    <default_database>default</default_database>
+
+    <mark_cache_size>5368709120</mark_cache_size>
+
+    <path>/var/lib/clickhouse/</path>
+    <tmp_path>/var/lib/clickhouse/tmp/</tmp_path>
+    <user_files_path>/var/lib/clickhouse/user_files/</user_files_path>
+    <access_control_path>/var/lib/clickhouse/access/</access_control_path>
+
+    <logger>
+        <level>debug</level>
+        <log>/var/log/clickhouse-server/clickhouse-server.log</log>
+        <errorlog>/var/log/clickhouse-server/clickhouse-server.err.log</errorlog>
+        <size>1000M</size>
+        <count>10</count>
+        <console>1</console>
+    </logger>
+
+</clickhouse>
--- a/.docker/clickhouse/users.xml
+++ b/.docker/clickhouse/users.xml
@@ -15,6 +15,7 @@
             </networks>
             <profile>default</profile>
             <quota>default</quota>
+            <access_management>1</access_management>
         </default>
     </users>
 
--- a/__tests__/integration/config.test.ts
+++ b/__tests__/integration/config.test.ts
@@ -14,7 +14,7 @@ describe('config', () => {
     await client.close()
   })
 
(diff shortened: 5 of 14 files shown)


__tests__/unit/format_query_settings.test.ts
```diff
@@ -13,12 +13,14 @@ describe('formatQuerySettings', () => {
...
+  it('formats a record', () => {
+    expect(formatQuerySettings({ foo: 'bar' })).toBe('{"foo":"bar"}')
+  })
+
   it('throws on unsupported values', () => {
     expect(() => formatQuerySettings(undefined as any)).toThrowError(
       'Unsupported value in query settings: [undefined].'
     )
-    expect(() => formatQuerySettings([1, 2] as any)).toThrowError(
-      'Unsupported value in query settings: [1,2].'
```
slvrtrn: an interesting one: as it is also an object, and we need to support "maps" (i.e. `Record<string, string>`), this does not throw as it passes the object check. I decided to not be that strict in this scenario and remove this test.

__tests__/unit/to_search_params.test.ts
```diff
@@ -16,12 +17,12 @@ describe('toSearchParams', () => {
...
       clickhouse_settings: {
         insert_quorum: 2,
-        distributed_product_mode: 'global',
+        distributed_product_mode: DistributedProductMode.GLOBAL,
         limit: 42,
       },
     })!
     expect(toSortedArray(params)).toEqual([
-      ['distributed_product_mode', 'global'],
+      ['distributed_product_mode', '2'],
```
slvrtrn: I will double-check if it indeed works as expected (i.e. numeric values instead of strings are recognized by the server)

src/client.ts
```diff
@@ -32,7 +31,9 @@ export interface ClickHouseClientConfigOptions {
 }
 
 export interface BaseParams {
-  clickhouse_settings?: ClickHouseSettings
+  clickhouse_settings?: ClickHouseSettings & {
+    wait_end_of_query?: 0 | 1
```
slvrtrn: as I understood this is an HTTP-only setting. Maybe there is a better place for it.

src/data_formatter/format_query_settings.ts
```diff
@@ -1,6 +1,9 @@
-export function formatQuerySettings(value: number | string | boolean): string {
+export function formatQuerySettings(
+  value: number | string | boolean | Record<string, string>
+): string {
   if (typeof value === 'boolean') return value ? '1' : '0'
   if (typeof value === 'number') return String(value)
   if (typeof value === 'string') return value
+  if (typeof value === 'object') return JSON.stringify(value)
```
slvrtrn: required to support CH "Map" type (used by `additional_table_filters` [that was introduced in 2022](https://clickhouse.com/docs/ru/whats-new/changelog/#new-feature))

__tests__/unit/to_search_params.test.ts
slvrtrn: OK, it's good that I double-checked... I actually executed every single enum type setting and fixed the typings.
I think we should compose an integration test as a follow-up.

src/settings.ts
```diff
@@ -0,0 +1,1365 @@
...
+  // Include column names in INSERT query (default: true)
+  output_format_sql_insert_include_column_names?: Bool
+  // Use REPLACE statement instead of INSERT (default: false)
+  output_format_sql_insert_use_replace?: Bool
+  // Quote column names with '`' characters (default: true)
+  output_format_sql_insert_quote_names?: Bool
+}
+
+type Bool = 0 | 1
+type Int64 = number | string
```
mshustov: Question: What if we use just `string` the types to avoid loss of precision for numbers > `Number.MAX_SAFE_INTEGER`?

src/settings.ts
```diff
@@ -0,0 +1,1365 @@
+export type ClickHouseSettings = {
```
mshustov: Could you add a comment on how you generated the file?

src/connection/connection.ts
```diff
@@ -1,7 +1,7 @@
 import Stream from 'stream'
-import type { ClickHouseSettings } from '../clickhouse_types'
 import type { Logger } from '../logger'
 import { HttpAdapter, HttpsAdapter } from './adapter'
+import { ClickHouseSettings } from '../settings'
```
mshustov: optional nit: it's used as a `type`. We can enable https://typescript-eslint.io/rules/consistent-type-imports in a separate PR.

src/data_formatter/format_query_settings.ts
```diff
@@ -1,6 +1,14 @@
...
+  value: number | string | boolean | Record<string, string>
+): string {
   if (typeof value === 'boolean') return value ? '1' : '0'
   if (typeof value === 'number') return String(value)
   if (typeof value === 'string') return value
+  // ClickHouse requires a specific, non-JSON format for passing maps
+  // as a setting value - single quotes instead of double
+  // Example: {'system.numbers':'number != 3'}
+  if (typeof value === 'object') {
+    return JSON.stringify(value).replace(/"/g, `'`)
```
mshustov: Nit: It breaks values containing `"`. It's an edge case, such users can workaround by providing a stringified value:
``{'foo':'b"ar'}``, I'm okay to merge as is.

src/client.ts
mshustov: Agreed, it should be separated from ClickHouseSettings. Maybe we place it in a dedicated file? There are more HTTP params to add https://github.com/ClickHouse/ClickHouse/blob/73709b04884ed7d228ac87f8a75a4a8c096ae598/src/Server/HTTPHandler.cpp#L668-L669

Maybe we should declare them as 
```typescript
interface ClickHouseServerSettings {
   // The actual size of the block to compress (default: 65536)
  min_compress_block_size?: UInt64
  ...
}
interface ClickHouseHttpSettings {
   wait_end_of_query?: Bool
   ...
}

interface ClickHouseSettings extends ClickHouseServerSetting, ClickHouseHttpSettings {}
```

src/settings.ts
```diff
@@ -0,0 +1,1365 @@
...
+  // The actual size of the block to compress (default: 65536)
```
mshustov: nit: it's not valid [TSDoc](https://tsdoc.org/). If we refactor the comment to a multi-line format to make valid TSDoc. 
Compare the autosuggest popup in VSCode:
- with an inline comment `// The actual size of the block to compress (default: 65536)`

- with a multi-line comment `/** The actual size of the block to compress (default: 65536) */`

 

WebStorm handles both types just fine:

src/settings.ts
```diff
@@ -0,0 +1,1365 @@
...
+  min_compress_block_size?: UInt64
+  // The maximum size of blocks of uncompressed data before compressing for writing to a table. (default: 1048576)
+  max_compress_block_size?: UInt64
+  // Maximum block size for reading (default: DEFAULT_BLOCK_SIZE)
```
mshustov: Did you keep it on purpose? https://github.com/ClickHouse/ClickHouse/blob/5f84f06d6d26672da3d97d0b236ebb46b5080989/src/Core/Defines.h#L20

src/settings.ts
```diff
@@ -0,0 +1,1365 @@
...
+  output_format_sql_insert_table_name?: string
+  // Include column names in INSERT query (default: true)
+  output_format_sql_insert_include_column_names?: Bool
+  // Use REPLACE statement instead of INSERT (default: false)
+  output_format_sql_insert_use_replace?: Bool
+  // Quote column names with '`' characters (default: true)
+  output_format_sql_insert_quote_names?: Bool
+}
+
+type Bool = 0 | 1
```
mshustov, line 1203: It's also discussable whether we need to handle them as `number` or `string`. We don't work with them as numbers. Even worse, every client handles them differently:
Python [as string](https://github.com/ClickHouse/clickhouse-connect/blob/1c5f375b494abcef3ddf2e17a4e7e64b9036d13e/clickhouse_connect/driver/httpclient.py#L136)
golang [as number](https://github.com/ClickHouse/clickhouse-go/blob/c54e9e192c46b1905336cd47464c39eb63482d71/tests/geo_multipolygon_test.go#L45)
jdbc [as boolean](https://github.com/ClickHouse/clickhouse-jdbc/blob/5770ae4dedb1f8a780dfa3e8abba737c3d611053/clickhouse-jdbc/src/main/java/ru/yandex/clickhouse/settings/ClickHouseQueryParam.java#L48)

We should discuss it within the group and align behavior across the clients

src/data_formatter/format_query_settings.ts
slvrtrn: I think there will be no nested values in this filter. It is better to use `Object.keys(...).map(...).join(..)` approach then.

src/settings.ts
slvrtrn: this is a good one, thanks

src/settings.ts
slvrtrn: it is a leftover. Will fix.

src/client.ts
slvrtrn: I've added a new issue for HTTP settings: https://github.com/ClickHouse/clickhouse-js/issues/53

PR #67, opened 2022-08-24
Throw an error when Rows.json/text/asStream are called twice

slvrtrn:


src/rows.ts
```diff
@@ -67,3 +72,31 @@ export class Rows {
...
+   */
+  json<T>(): T {
+    if (this._json === undefined) {
+      this._json = decode(this.text(), this.format)
+    }
+    return this._json as T
+  }
+}
+
+const StreamAlreadyConsumed = 'Stream has been already consumed'
```
mshustov: Nit: it's not a factory or a type. Let's start with a char in lower case
`StreamAlreadyConsumed --> streamAlreadyConsumedMessage`

src/rows.ts
```diff
@@ -67,3 +72,31 @@ export class Rows {
...
+   * Returns a JSON representation of a row.
+   * The method will throw if called on a response in JSON incompatible format.
+   *
+   * It is safe to call this method multiple times as the result is cached.
+   */
+  json<T>(): T {
+    if (this._json === undefined) {
+      this._json = decode(this.text(), this.format)
+    }
+    return this._json as T
```
mshustov: What are the benefits of adding a local state? It's not likely someone calls `json` a few times

__tests__/unit/rows.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+describe('rows', () => {
+  const row1 = '{"foo":"bar"}\n'
+  const row2 = '{"qaz":"qux"}\n'
+  const expectedText = `{"foo":"bar"}\n{"qaz":"qux"}\n`
+  const expectedJson = [{ foo: 'bar' }, { qaz: 'qux' }]
+
+  const err = 'Stream has been already consumed'
+
+  it('should consume the response as text, then as JSON', async () => {
```
mshustov: It's not the logic this suite tests.
`should consume the response as text only once`

__tests__/unit/rows.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    const rows = makeRows()
+
+    expect(await rows.text()).toEqual(expectedText)
+    await expect(rows.text()).rejects.toThrowError(err)
+    await expect(rows.json()).rejects.toThrowError(err)
+  })
+
+  it('should consume the response as JSON, then as text', async () => {
```
mshustov: the same problem with the description here

__tests__/unit/rows.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    const rows = makeRows()
+
+    expect(await rows.json()).toEqual(expectedJson)
+    await expect(rows.json()).rejects.toThrowError(err)
+    await expect(rows.text()).rejects.toThrowError(err)
+  })
+
+  it('should consume the response as an arbitrary stream', async () => {
```
mshustov: `should consume the response as a stream of Row instances`

__tests__/unit/rows.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    expect(result).toEqual(expectedJson)
+    expect(stream.readableEnded).toBeTruthy()
+
+    expect(() => rows.asStream()).toThrowError(err)
+    await expect(rows.json()).rejects.toThrowError(err)
+    await expect(rows.text()).rejects.toThrowError(err)
+  })
+
+  function makeRows() {
```
mshustov, line 55: Shouldn't we move it closed to `row1` and `row2` declaration?

__tests__/unit/rows.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  it('should consume the response as an arbitrary stream', async () => {
+    const rows = makeRows()
+    const stream = rows.asStream()
+
+    expect(stream.readableEnded).toBeFalsy()
+
+    const result = []
+    for await (const row of stream) {
+      result.push((row as Row).json())
```
mshustov, line 34: Should we add a test that `Row.text()` and `Row.json()` can be called several times?

slvrtrn: Let's do that

src/rows.ts
slvrtrn: fair enough. Let's remove

PR #68, opened 2022-08-24
Support CSV/TSV/CustomSeparated raw streaming

slvrtrn:
Partially implements #28 

Adds 
```
  'CSV',
  'CSVWithNames',
  'CSVWithNamesAndTypes',
  'TabSeparated',
  'TabSeparatedRaw',
  'TabSeparatedWithNames',
  'TabSeparatedWithNamesAndTypes',
  'CustomSeparated',
  'CustomSeparatedWithNames',
  'CustomSeparatedWithNamesAndTypes',
```
as possible raw streaming formats

### DIFF
--- a/__tests__/unit/client.test.ts
+++ b/__tests__/unit/client.test.ts
@@ -69,6 +69,12 @@ describe('validateInsertValues', () => {
       'CSVWithNames',
       'CSVWithNamesAndTypes',
       'TabSeparated',
+      'TabSeparatedRaw',
+      'TabSeparatedWithNames',
+      'TabSeparatedWithNamesAndTypes',
+      'CustomSeparated',
+      'CustomSeparatedWithNames',
+      'CustomSeparatedWithNamesAndTypes',
     ]
     rawFormats.forEach((format) => {
       expect(() =>


src/client.ts
```diff
@@ -199,14 +199,19 @@ function formatCommandQuery(query: string, format: DataFormat | false): string {
...
 ): void {
   if (Array.isArray(values) === false && isStream(values) === false) {
     throw new Error(
       'Insert expected "values" to be an array or a stream of values.'
     )
   }
 
+  if (format.startsWith('CSV')) {
+    return
+  }
```
slvrtrn: so it works just fine with these two short-circuits (this one and in the `formatter.ts`) as it seems. We can leave it as is or think about a better insert values abstraction as discussed.

mshustov: The current approach doesn't allow inserting `JSON*` formats as strings/bytes. Let's imagine a user reads a `ndjson` file to insert into DB. Do they need to parse it to convert it into JSON now?

__tests__/integration/insert_csv_file_stream.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    await client.insert({
+      table: tableName,
+      values: stream,
+      format: 'CSVWithNamesAndTypes',
+    })
+    await assertInsertedValues()
+  })
+
+  it('should throw in case of invalid CSV format', async () => {
+    const stream = createFileStream('insert_csv_invalid_format.csv')
```
mshustov: Do we need to create files? A Readable stream can be instantiated with a list of strings
```js
Stream.from(['foobar,42,,'])
```

__tests__/integration/insert_stream_raw_formats.test.ts
```diff
@@ -0,0 +1,241 @@
+import { createTestClient, guid } from '../utils'
+import type { ClickHouseClient, ClickHouseSettings } from '../../src'
+import { createSimpleTable } from './fixtures/simple_table'
+import Stream from 'stream'
+
+// FIXME `npm run test -- --detectOpenHandles`
+//  Jest has detected the following 13 open handles
+//  potentially keeping Jest from exiting
```
slvrtrn, line 8: Still a mystery why this is happening

src/data_formatter/formatter.ts
```diff
@@ -80,6 +86,12 @@ export function decode(text: string, format: DataFormat): any {
  * @returns string
  */
 export function encode(value: any, format: DataFormat): string {
+  // TODO: add
+  //   'JSONStringsEachRow',
+  //   'JSONCompactEachRowWithNames',
+  //   'JSONCompactEachRowWithNamesAndTypes',
+  //   'JSONCompactStringsEachRowWithNames',
+  //   'JSONCompactStringsEachRowWithNamesAndTypes',
```
slvrtrn, line 94: I guess support for these formats should also go into this `encode` function as a follow-up, WDYT @mshustov?

Issue #57, opened 2022-08-19
Memory leaks tests


slvrtrn:
- [x] File streaming test #71 
- [x] Random data with basic data types streaming #71 
- [x] Random arrays
- [x] Random data with complex (Array of Arrays, Maps) data types streaming

PR #71, opened 2022-08-26
Memory leak benchmarks (file and random integers streaming)

slvrtrn:
Related to #57

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -11,7 +11,7 @@ module.exports = {
   testTimeout: 30000,
   globals: {
     'ts-jest': {
-      tsconfig: 'tsconfig.eslint.json',
+      tsconfig: 'tsconfig.dev.json',
     },
   },
 }


.gitignore
```diff
@@ -1,4 +1,8 @@
 package-lock.json
 dist/
 .idea
-node_modules
\ No newline at end of file
+node_modules
+lib
```
mshustov: Maybe we should output into `build` folder?

benchmarks/leaks/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+```sh
+docker-compose up -d
+```
+
+#### Prepare input data
+
+```sh
+mkdir -p benchmarks/leaks/input \
+&& wget https://datasets.clickhouse.com/mgbench1.csv.xz \
```
mshustov: Why not curl? Not every person has `wget` install on the local env.
The same about `xz`.

benchmarks/leaks/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+```sh
+tsc --project benchmarks/tsconfig.benchmarks.json \
+&& node --expose-gc benchmarks/build/benchmarks/leaks/memory_leak_brown.js
+```
+
+
+## Random data streaming test
+
+---
```
mshustov, line 49: Does it mean `TODO`?

src/client.ts
```diff
@@ -142,6 +142,7 @@ export class ClickHouseClient {
     return new Rows(stream, format)
   }
 
+  // FIXME: allow ; in the end of the query
```
mshustov, line 145: What does it mean?

benchmarks/leaks/README.md
```diff
@@ -0,0 +1,49 @@
...
+&& mv mgbench1.csv benchmarks/leaks/input \
+&& rm mgbench1.csv.xz
+```
+
+See [official examples](https://clickhouse.com/docs/en/getting-started/example-datasets/brown-benchmark/) for more information.
+
+#### Run the test
+
+```sh
+tsc --project benchmarks/tsconfig.benchmarks.json \
```
mshustov: Why don't you use `tsconfig.eslint.json` (`npm run build:all`)?
And we need to rename `tsconfig.eslint.json`, for sure.

benchmarks/leaks/shared.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export interface MemoryUsage {
+  rss: number
+  heapTotal: number
+  heapUsed: number
+  external: number
+  arrayBuffers: number
+}
+
+export function getMemoryUsageInMegabytes(print = true): MemoryUsage {
```
mshustov: Why not separate print from formatting?

benchmarks/leaks/README.md
slvrtrn: That's from the [docs](https://clickhouse.com/docs/en/getting-started/example-datasets/brown-benchmark/)

benchmarks/leaks/README.md
slvrtrn: I committed the second part :)

src/client.ts
slvrtrn: we need to allow queries that end with a semicolon. Currently, `SELECT * FROM table;` will fail, logging that "multiline statements are not allowed", cause the client will try to attach `FORMAT` after the semicolon.

slvrtrn: I will fix it in a follow-up PR

.gitignore
slvrtrn: This is actually unused now, I will remove this from `gitignore`

benchmarks/leaks/README.md
slvrtrn: A good one, I will try.

benchmarks/leaks/README.md
slvrtrn: Unfortunately, `xz-utils` are required here. Could not do it with just `tar`

PR #72, opened 2022-08-29
Allow queries with semicolon in the end

slvrtrn:
Previously, something like `client.select({ query: 'SELECT * FROM system.numbers LIMIT 5;' })` would've failed because format was added after the trailing semi.

### DIFF
--- a/__tests__/integration/command.test.ts
+++ b/__tests__/integration/command.test.ts
@@ -83,6 +83,14 @@ describe('command', () => {
       expect(await numbers.text()).toEqual('0\n1\n2\n')
     })
 
+    it('should cut off the statements after the first semi', async () => {
+      const numbers = await client.command({
+        query: 'SELECT * FROM system.numbers LIMIT 3;asdf foobar',
+        format: 'CSV',
+      })
+      expect(await numbers.text()).toEqual('0\n1\n2\n')
+    })
+
     it('should allow commands without format but with trailing semi', async () => {
       const result = await client.command({
         query: 'EXISTS system.databases;',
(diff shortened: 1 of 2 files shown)


__tests__/integration/command.test.ts
```diff
@@ -66,6 +66,40 @@ describe('command', () => {
     )
   })
 
+  describe('trailing semi', () => {
+    it('should allow queries with trailing semi', async () => {
+      const numbers = await client.command({
+        query: 'SELECT * FROM system.numbers LIMIT 3;',
```
slvrtrn, line 72: since it is still a perfectly valid command, decided to just use SELECT here.

__tests__/integration/select.test.ts
```diff
@@ -109,9 +109,9 @@ describe('select', () => {
   })
 
   it('returns an error details provided by ClickHouse', async () => {
-    await expect(client.select({ query: ';' })).rejects.toEqual(
+    await expect(client.select({ query: 'foobar' })).rejects.toEqual(
```
slvrtrn, line 112: Interesting how this test was green before, cause the resulting query was `;\nFORMAT JSON`

image

src/client.ts
```diff
@@ -188,17 +187,27 @@ export function validateSelectQuery(query: string): void {
...
 function formatCommandQuery(query: string, format: DataFormat | false): string {
   query = query.trim()
   if (format !== false) {
+    query = removeSemi(query)
     return query + ' \nFORMAT ' + format
   }
   return query
 }
 
+function removeSemi(query: string) {
```
mshustov, line 203: It might trim some statements after `;`. 
ClickHouse client allows using `;` as a query separator in `insert`. see https://github.com/ClickHouse/ClickHouse/issues/12288 and https://github.com/ClickHouse/ClickHouse/pull/13192/

IMO it's okay if the client doesn't support this functionality since users are recommended to use `insert` API instead. Let's just add a test to show it's a known limitation.

Issue #55, opened 2022-08-19
Additional integration tests


slvrtrn:
# Happy path

Common
--
- [x] Can create several instances on a language client #75 
- [x] Clients can send parallel requests (DDL, select, insert stream, insert array) #75  

Configuration
--
- [x] Can connect to ClickHouse with the default configuration
- [x] Can create an HTTPS connection
- [x] Can specify default database name on creation #76 
- [x] Can configure ClickHouse settings for every client request #76 
- [x] Can provide a custom logger class on creation #76 
- [x] Language client does not mutate provided configuration #76 

Insert & Select
--
- [x] Can perform several parallel requests #77
- [x] Cancellation of one of the parallel requests does not cancel the others #77
- [x] A query resulting in a single row
- [x] A query resulting in a dataset of X rows
- [x] A dataset of several rows with all the language primitives #78 
- [x] A test suit for each supported ClickHouse Data Type (Enum, Map, UInt 256, etc.) #78 
- [x] A stream of text data (TSV or ndjson)
- [x] A test suit for Insert and Select for chosen [supported format](https://clickhouse.com/docs/en/interfaces/formats/) #83 
- [x] Insert and Select a DateTime as a number, string, or Date object in a timezone different from the server's one #82 
- [x] A result dataset containing `toDateTime` conversion to a time zone different from the client's one #82 
- [x] Can send a query with parameters
- [x] Can send a query with non-ASCII symbols #80 
- [x] Can send a query with parameters containing non-ASCII symbols #80 
- [x] Can send a query with parameters containing DateTime / Date #80 
- [x] Can send a query with enums in parameters #80 
- [x] A query with a response summary
- [x] A query with query-specific settings taking precedence over instance-specific settings
- [x] Accepts a compressed response
- [x] Sends a compressed request
- [x] Can send a multiline query
- [x] Can send a query with an inline comment
- [x] Can send a query with a multiline comment
- [x] Can send a query with a comment on a trailing line
- [x] Ignores unknown properties during `insert` https://github.com/ClickHouse/clickhouse-js/pull/83

Insert
--
- [x] A stream of text data (TSV, CSV)
- [x] A stream of data objects

Select
--
- [x] A query that results in an empty response #79
- [x] Sends a stream of data (TSV or ndjson)
- [x] A query with a response in [JSON format (with JSON default settings)](https://clickhouse.com/docs/en/interfaces/formats/#json-formats-settings) 
[...]

PR #75, opened 2022-09-01
Multiple client instances integration tests

slvrtrn:
#55 

Common
--
- [x] Can create several instances on a language client
- [x] Clients can send parallel requests (DDL, select, insert stream, insert array)

### DIFF
--- a/__tests__/integration/multiple_clients.test.ts
+++ b/__tests__/integration/multiple_clients.test.ts
@@ -1,0 +1,123 @@
+import type { ClickHouseClient } from '../../src'
+import { createSimpleTable } from './fixtures/simple_table'
+import { createTestClient, guid } from '../utils'
+import Stream from 'stream'
+
+const CLIENTS_COUNT = 5
+
+describe('multiple clients', () => {
+  const clients: ClickHouseClient[] = Array(CLIENTS_COUNT)
+
+  beforeEach(() => {
+    for (let i = 0; i < CLIENTS_COUNT; i++) {
+      clients[i] = createTestClient()
+    }
+  })
+
+  afterEach(async () => {
+    for (const c of clients) {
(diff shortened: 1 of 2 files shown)


__tests__/integration/multiple_clients.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      clients.map((client, i) =>
+        client
+          .select({
+            query: `SELECT * FROM system.numbers LIMIT ${i + 1}`,
+            format: 'JSONEachRow',
+          })
+          .then(async (r) => {
+            results[i] = (await r.json<Res>())
+              .map((r) => +r.number)
+              .reduce((prev, cur) => prev + cur)
```
mshustov, line 36: Isn't asserting a result it received enough? What does `reduce` give us?

__tests__/integration/multiple_clients.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    }
+  })
+
+  afterEach(async () => {
+    for (const c of clients) {
+      await c.close()
+    }
+  })
+
+  it('should send multiple parallel selects', async () => {
```
mshustov, line 23: we can add a test that if a request is aborted with `signal`, it doesn't affect other parallel requests

__tests__/integration/multiple_clients.test.ts
slvrtrn: just to have a single number as a result. It was handy when I tested it with 20 client instances causing result sets to become really lengthy.

__tests__/integration/multiple_clients.test.ts
slvrtrn: I will add this test in the scope of the "single client - multiple requests" test

PR #77, opened 2022-09-02
Simultaneous requests tests (single client)

slvrtrn:
#55 

Insert & Select
---
- [x] Non-streaming cases
- [x] Streaming cases
- [x] Abort one request while keeping others

### DIFF
--- a/__tests__/integration/abort_request.test.ts
+++ b/__tests__/integration/abort_request.test.ts
@@ -1,12 +1,9 @@
 import { AbortController } from 'node-abort-controller'
 import { type ClickHouseClient, type ResponseJSON } from '../../src'
-import {
-  createTable,
-  createTestClient,
-  guid,
-  makeObjectStream,
-  TestEnv,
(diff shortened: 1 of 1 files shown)


__tests__/integration/multiple_clients.test.ts
```diff
@@ -21,29 +21,20 @@ describe('multiple clients', () => {
...
-    type Res = Array<{ number: string }>
-    const results: Record<string, number> = {}
+    type Res = Array<{ sum: number }>
+    const results: number[] = []
     await Promise.all(
       clients.map((client, i) =>
         client
           .select({
-            query: `SELECT * FROM system.numbers LIMIT ${i + 1}`,
+            query: `SELECT toInt32(sum(*)) AS sum FROM numbers(0, ${i + 2});`,
```
slvrtrn, line 30: Simplified this one as a follow-up for #75 as it was indeed a bit complicated and we can perfectly do that on the database side.

PR #78, opened 2022-09-02
Test all common and complex data types (minus Nested)

slvrtrn:
#55 

Insert & Select
--
- [x] A dataset of several rows with all the language primitives
- [x] A test suit for each supported ClickHouse Data Type (Enum, Map, UInt 256, etc.)

`Nested` is skipped for the time being. 
`Decimal` does not work with the `JSON*` input format, not sure if it is intended or not.

### DIFF
--- a/__tests__/integration/data_types.test.ts
+++ b/__tests__/integration/data_types.test.ts
@@ -1,0 +1,488 @@
+import type { ClickHouseClient } from '../../src'
+import { createTable, createTestClient, guid, TestEnv } from '../utils'
+import { v4 } from 'uuid'
+import { randomInt } from 'crypto'
+import Stream from 'stream'
+
+describe('data types', () => {
+  let client: ClickHouseClient
+  beforeEach(() => {
+    client = createTestClient()
+  })
+  afterEach(async () => {
+    await client.close()
+  })
+
+  describe('primitives', () => {
+    it('should work with integer types', async () => {
+      const values = [
+        {
+          i1: 127,
+          i2: 32767,
+          i3: 2147483647,
+          i4: '9223372036854775807',
+          i5: '170141183460469231731687303715884105727',
+          i6: '57896044618658097711785492504343953926634992332820282019728792003956564819967',
+          u1: 255,
+          u2: 65535,
+          u3: 4294967295,
+          u4: '18446744073709551615',
+          u5: '340282366920938463463374607431768211455',
+          u6: '115792089237316195423570985008687907853269984665640564039457584007913129639935',
+        },
+        {
+          i1: -128,
+          i2: -32768,
+          i3: -2147483648,
+          i4: '-9223372036854775808',
+          i5: '-170141183460469231731687303715884105728',
+          i6: '-57896044618658097711785492504343953926634992332820282019728792003956564819968',
+          u1: 120,
+          u2: 1234,
+          u3: 51234,
+          u4: '421342',
+          u5: '15324355',
+          u6: '41345135123432',
+        },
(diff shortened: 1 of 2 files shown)


__tests__/integration/schema_types.test.ts
```diff
@@ -383,12 +383,6 @@ async function assertInsertAndSelect<S extends ch.Shape>(
   await table.insert({
     values: value,
   })
-  const result = await (
-    await table.select({
-      clickhouse_settings: {
-        s3_create_new_file_on_insert: 1,
```
slvrtrn, line 389: `¯\_(ツ)_/¯`

__tests__/integration/data_types.test.ts
```diff
@@ -0,0 +1,488 @@
...
+      {
+        'n.id': 43,
+        'n.name': 'bar',
+        'n.createdAt': '2000-01-12',
+        'n.roles': ['Admin'],
+      },
+    ])
+  })
+
+  // TODO should belong to fixtures as it is quite useful in other tests
```
slvrtrn, line 438: will move in the next one

__tests__/integration/data_types.test.ts
```diff
@@ -0,0 +1,488 @@
...
+  //  -     "n.name": "foo",
+  //  -     "n.roles": Array [
+  //  -       "User",
+  //  -     ],
+  //  +     "n.createdAt": Array [],
+  //  +     "n.id": Array [],
+  //  +     "n.name": Array [],
+  //  +     "n.roles": Array [],
+  //    },
+  it.skip('should work with nested', async () => {
```
slvrtrn, line 386: I genuinely have no idea how to insert and query it properly after spending some time on the test.
Neither nested objects like below nor fields like `n.id` worked out. Noticed an interesting thing though: to get field names like `n.id` to work, you need to manually escape it:

```
{
  '`n.id`': 42,
  '`n.name`': 'foo',
  '`n.createdAt`': '2001-04-23',
  '`n.roles`': ['User'],
}
```
as our HTTP interface does not like the identifiers with dots.

__tests__/integration/data_types.test.ts
```diff
@@ -0,0 +1,488 @@
...
+    // FIXME (?) does not work with JSON* insert formats, only raw streaming
+    //   as decimals are required to be passed as numbers
+    //   when using JSONEachRow (why?)
+    it('should work with decimals', async () => {
+      const stream = new Stream.Readable({
+        objectMode: false,
+        read() {
+          //
+        },
+      })
```
slvrtrn, line 105: Stream is required to bypass the `encode` function which only works with `JSON*` currently.
This is a very tricky case.

https://github.com/ClickHouse/clickhouse-js/blob/7e29e4f4c6b85923eec588e828fbb897bcfec41e/src/data_formatter/formatter.ts#L88-L99

__tests__/integration/data_types.test.ts
```diff
@@ -0,0 +1,488 @@
...
+      ]
+      const table = await createTableWithFields('ip1 IPv4, ip2 IPv6')
+      await insertAndAssert(table, values)
+    })
+
+    it('should work with ((very) nested) arrays', async () => {
+      // it's the largest reasonable nesting value (data is generated within 50 ms);
+      // 25 here can already tank the performance to ~500ms only to generate the data;
+      // 50 simply times out :)
+      const maxNestingLevel = 20
```
slvrtrn, line 234: I could not resist testing such an edge case (also for maps)

PR #80, opened 2022-09-03
Various parametrized queries tests, fix query params millis

slvrtrn:
#55

Insert & Select
--
- [x] Can send a query with non-ASCII symbols
- [x] Can send a query with parameters containing non-ASCII symbols
- [x] Can send a query with parameters containing DateTime / Date
- [x] Can send a query with enums in parameters

Command
--
- [x] Can send a query with parameters

### DIFF
--- a/__tests__/unit/format_query_params.test.ts
+++ b/__tests__/unit/format_query_params.test.ts
@@ -3,7 +3,18 @@ import { formatQueryParams } from '../../src/data_formatter'
 // JS always creates Date object in local timezone,
 // so we might need to convert the date to another timezone
 function convertDateToTimezone(date: Date, tz: string) {
-  return new Date(date.toLocaleString('en-US', { timeZone: tz }))
+  return new Date(
+    date.toLocaleString('en-US', {
+      timeZone: tz,
+      year: 'numeric',
+      month: 'numeric',
+      day: 'numeric',
+      hour: 'numeric',
+      minute: 'numeric',
+      second: 'numeric',
+      fractionalSecondDigits: 3, // print millis
+    })
+  )
 }
 
 describe('formatQueryParams', () => {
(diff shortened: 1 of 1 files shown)


__tests__/integration/select_query_binding.test.ts
```diff
@@ -128,4 +182,71 @@ describe('select with query binding', () => {
...
+      expect(response).toBe('2\n')
+    })
+
+    it('should accept numeric enums in a parametrized query', async () => {
+      enum MyEnum {
+        foo = 'foo',
+        bar = 'bar',
+      }
+      const rows = await client.select({
+        query: 'SELECT concat({str1: String},{str2: String})',
```
mshustov, line 226: Is it possible to declare an expected type as `Enum` instead of `String`?

src/data_formatter/format_query_params.ts
```diff
@@ -34,7 +48,9 @@ export function formatQueryParams(
     const time = `${withPadding(value.getHours())}:${withPadding(
       value.getMinutes()
     )}:${withPadding(value.getSeconds())}`
-    return `${date} ${time}`
+
+    const ms = formatMillis(value)
+    return `${date} ${time}${ms}`
```
mshustov, line 53: nit: It could be tested in a unit test

slvrtrn: not a nit, that's a good one! 👍

__tests__/integration/select_query_binding.test.ts
slvrtrn: let's see. Maybe we can do some cast shenanigans like in one of the other tests.

slvrtrn: Seems to be quite complicated. Let's skip this one.

PR #82, opened 2022-09-04
DateTime and various unhappy path tests

slvrtrn:
#55 

NB: I'd suggest we don't use `Date` as an available input type for object insertions for the time being because of this one:

https://github.com/ClickHouse/clickhouse-js/blob/139168126c023a647e73d835a85863bab6086916/src/data_formatter/formatter.ts#L95-L97

`JSON.stringify` formats it as 

```js
JSON.stringify(new Date())
'"2022-09-04T22:38:07.082Z"'
```

and ClickHouse does not like this format.

Regarding the "Provides details of a network error" test case: for now, `ECONNREFUSED` is covered when we query an unreachable endpoint, but we can try to simulate a disconnect in the middle of the execution as a follow-up

---

Insert & Select
--
- [x] Insert and Select a DateTime as a number, string, ~~or Date object~~ in a timezone different from the server's one
- [x] A result dataset containing `toDateTime` conversion to a time zone different from the client's one

Insert & select
--
- [x] Provides error details when sending a query with unknown setting/parameters
- [x] Provides details of a network error

Insert
--
- [x] Provides error details on an exception raised during Insert (a dataset with an invalid type)

### DIFF
--- a/__tests__/integration/date_time.test.ts
+++ b/__tests__/integration/date_time.test.ts
@@ -33,6 +33,28 @@ describe('DateTime', () => {
     })
   })
 
+  describe('Date32', () => {
+    it('should insert Date32 and get it back', async () => {
+      // currently, there is no way to insert a Date32 as a number via HTTP
+      // the conversion is not performed automatically like in VALUES clause
+      const table = await createTableWithFields(client, 'd Date32')
+      await client.insert({
+        table,
+        values: [{ d: '2022-09-05' }],
+        format: 'JSONEachRow',
+      })
+
+      expect(
+        await client
+          .select({
+            query: `SELECT * EXCEPT id FROM ${table}`,
+            format: 'JSONEachRow',
+          })
+          .then((r) => r.json())
+      ).toEqual([{ d: '2022-09-05' }])
+    })
+  })
+
   describe('DateTime', () => {
     it('should insert DateTime and get it back', async () => {
       const table = await createTableWithFields(client, 'd DateTime')
--- a/__tests__/integration/error_parsing.test.ts
+++ b/__tests__/integration/error_parsing.test.ts
@@ -83,9 +83,8 @@ describe('error', () => {
       })
     ).rejects.toEqual(
       expect.objectContaining({
+        message: 'connect ECONNREFUSED 127.0.0.1:1111',
         code: 'ECONNREFUSED',
-        address: '127.0.0.1',
-        port: 1111,
       })
     )
   })


__tests__/integration/data_types.test.ts
```diff
@@ -1,8 +1,9 @@
 import type { ClickHouseClient } from '../../src'
```
slvrtrn, line 1: For this file, that's just reformating (I removed unnecessary `describe` clause) + `createTableWithFields` is moved to fixtures now
Please ignore

__tests__/integration/error_parsing.test.ts
```diff
@@ -71,4 +71,22 @@ describe('error', () => {
...
+      host: 'http://localhost:1111',
+    })
+    await expect(
+      client.select({
+        query: 'SELECT * FROM system.numbers LIMIT 3',
+      })
+    ).rejects.toEqual(
+      expect.objectContaining({
+        code: 'ECONNREFUSED',
+        address: '127.0.0.1',
```
mshustov: That's not a standard `Error` interface. Maybe we should assert against well-known properties? `name`, `message`, maybe even `code` since ClickHouseError provides it.

__tests__/integration/select_query_binding.test.ts
```diff
@@ -248,5 +248,24 @@ describe('select with query binding', () => {
       const response = await rows.text()
       expect(response).toBe(`Enum8(\\'a\\' = 1, \\'b\\' = 2)\n`)
     })
+
+    it('should provide error details when sending a request with an unknown parameter', async () => {
+      await expect(
+        client.select({
+          query: `
+            SELECT * FROM system.numbers 
+            WHERE number > {min_limit: UInt64} LIMIT 3
```
mshustov, line 257: What happens if code provides an excessive parameter?

__tests__/integration/date_time.test.ts
```diff
@@ -0,0 +1,191 @@
...
+  })
+
+  describe('Date', () => {
+    it('should insert Date and get it back', async () => {
+      // currently, there is no way to insert a Date as a number via HTTP
+      // the conversion is not performed automatically like in VALUES clause
+      const table = await createTableWithFields(client, 'd Date')
+      await client.insert({
+        table,
+        values: [{ d: '2022-09-05' }],
```
mshustov, line 21: `Date` can be inserted as a number as well

__tests__/integration/date_time.test.ts
```diff
@@ -0,0 +1,191 @@
...
+describe('DateTime', () => {
+  let client: ClickHouseClient
+  beforeEach(() => {
+    client = createTestClient()
+  })
+  afterEach(async () => {
+    await client.close()
+  })
+
+  describe('Date', () => {
```
mshustov, line 14: Could you add a test for `Date32`?

__tests__/integration/date_time.test.ts
slvrtrn: Unfortunately, is not possible via HTTP at the moment.

__tests__/integration/select_query_binding.test.ts
slvrtrn: it's just ignored

PR #86, opened 2022-09-13
Unify `select` and `command`, introduce raw `exec`

slvrtrn:
* `select` + some `command` cases -> `query`
* `command` where the output is mostly irrelevant (DDLs etc) -> `exec`

### DIFF
--- a/__tests__/utils/client.ts
+++ b/__tests__/utils/client.ts
@@ -21,11 +21,8 @@ export function createTestClient(
     }`
   )
   const clickHouseSettings: ClickHouseSettings = {}
-  if (env === TestEnv.LocalCluster) {
+  if (env === TestEnv.LocalCluster || env === TestEnv.Cloud) {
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -133,29 +128,22 @@ export class ClickHouseClient {
...
-
     return new Rows(stream, format)
   }
 
-  async command(params: CommandParams): Promise<Rows> {
-    const format = params.format === undefined ? 'JSON' : params.format
-    const query = formatCommandQuery(params.query, format)
-
-    const stream = await this.connection.command({
+  exec(params: ExecParams): Promise<Stream.Readable> {
```
slvrtrn, line 141: Since it is just raw command execution, I decided not to wrap it with `Rows` as it also requires `format` which we don't provide here. 

However, for ergonomics, it'd be great to have the `getAsText` utility function available in the library exports I think. (see how it's used in the tests)

PR #88, opened 2022-09-13
Update README, `Rows.asStream()` -> `Rows.stream()`, remove FIXME

### LINKED ISSUE #29: Provide documentation
Documentation (README and the company website)
- Requirements for the environment
- Installation instructions
- API usage examples (getting started):
    - Instantiation of the client
    - Create a table
    - Write to ClickHouse
    - Read from ClickHouse
    - Configuration options
- Provide recommended usage patterns and caveats
- A link to the ClickHouse version compatibility matrix on the official website.

slvrtrn:
Resolves #29

### DIFF
--- a/__tests__/utils/retry.test.ts
+++ b/__tests__/utils/retry.test.ts
@@ -16,10 +16,6 @@ describe('retryOnFailure', () => {
     let result = 0
     setTimeout(() => {
       result = 42
-      // FIXME: setTimeout has typings of window.setTimeout & globals.setTimeout
-      //  remove when DOM lib is excluded from typings.
-      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-      // @ts-ignore
     }, 1000).unref()
     await expect(
       retryOnFailure(


README.md
```diff
@@ -1,8 +1,368 @@
...
+<a href="https://github.com/ClickHouse/clickhouse-js/actions/workflows/run-tests.yml">
+<img src="https://github.com/ClickHouse/clickhouse-js/actions/workflows/run-tests.yml/badge.svg?branch=main">
+</a>
+</p>
 
-## Requirements
+## Introduction
 
-- Node.JS 14.x or 16.x
+Official Node.js (14.x or 16.x) client for [ClickHouse](https://clickhouse.com/), written purely in TypeScript, thoroughly tested with ClickHouse 22.8 and upcoming 22.9.
```
mshustov: Maybe instead of hardcoding the versions, we can create a single compatibility table to update versions in a single place?

README.md
```diff
@@ -1,8 +1,368 @@
...
+
+It is focused on data streaming for both inserts and selects using standard [Node.js Streaming API](https://nodejs.org/api/stream.html).
+
+As of now, HTTP(s) protocol is used under the hood, with [planned](https://github.com/ClickHouse/clickhouse-js/issues/25) Native protocol support.
```
mshustov: we need to create a new issue. `native protocol` != `native data format`

README.md
```diff
@@ -1,8 +1,368 @@
...
+</a>
+</p>
 
-## Requirements
+## Introduction
 
-- Node.JS 14.x or 16.x
+Official Node.js (14.x or 16.x) client for [ClickHouse](https://clickhouse.com/), written purely in TypeScript, thoroughly tested with ClickHouse 22.8 and upcoming 22.9.
+
+It is focused on data streaming for both inserts and selects using standard [Node.js Streaming API](https://nodejs.org/api/stream.html).
```
mshustov: Nit: the link points to `node v18` docs. This version is not supported by the client yet. We can start testing against node v18, even though Active LTS for v18 is starting from 25 Oct 2022 https://nodejs.org/en/about/releases/

README.md
```diff
@@ -1,8 +1,368 @@
...
+
+As of now, HTTP(s) protocol is used under the hood, with [planned](https://github.com/ClickHouse/clickhouse-js/issues/25) Native protocol support.
+
+## Connection
+
+A very basic connection to a local ClickHouse instance with default settings (for example, if it is running as a Docker container):
```
mshustov: Let's add a link to `CONTRIBUTING` where we describe how to run the ClickHouse Docker image locally.

README.md
mshustov: nit: I'd rather move it to `Connection` chapter

README.md
```diff
@@ -1,8 +1,368 @@
...
+
+  // milliseconds, default 10_000
+  connect_timeout?: number
+
+  // milliseconds, default 300_000
+  request_timeout?: number
+
+  // by default, Node.js HTTP(s) Agent has infinite max open sockets
+  // it can be overriden with this setting
+  max_open_connections?: number
```
mshustov, line 92: are you going to support it soon? [the link](https://github.com/ClickHouse/clickhouse-js/blob/60c484a3492420baed4b4c6c33cc0845262285e7/src/client.ts#L13-L35) points to code where it's commented out

README.md
```diff
@@ -1,8 +1,368 @@
...
+
+  // HTTP compression settings
+  compression?: {
+    // enabled by default
```
mshustov: Let's clarify it instructs ClickHouse to respond with the compressed response body

README.md
```diff
@@ -1,8 +1,368 @@
...
+  // if unset, defaults to http://localhost:8123
+  host?: string
+
+  // milliseconds, default 10_000
+  connect_timeout?: number
+
+  // milliseconds, default 300_000
+  request_timeout?: number
+
+  // by default, Node.js HTTP(s) Agent has infinite max open sockets
```
mshustov: Maybe we shouldn't mention *HTTP Agent* in public interfaces if we're going to support Native TCP protocol later.

README.md
```diff
@@ -1,8 +1,368 @@
...
+  database?: string
+
+  // additional settings to send with every query, such as `date_time_input_format` or `insert_quorum`
+  // see https://clickhouse.com/docs/en/operations/settings/settings/
+  // typings should support most of the options listed there
+  clickhouse_settings?: ClickHouseSettings
+
+  // logger settings
+  log?: {
+    // enabled by default, can be disabled
```
mshustov: It's disabled by default. Isn't it? https://github.com/ClickHouse/clickhouse-js/blob/main/src/client.ts#L110-L112

README.md
```diff
@@ -1,8 +1,368 @@
...
+- [ClickHouseSettings](https://github.com/ClickHouse/clickhouse-js/blob/730b1b2516e2d47dc9a32b1d8d0b8ba8ceb95ead/src/settings.ts#L10-L1201)
+
+- [Logger](https://github.com/ClickHouse/clickhouse-js/blob/ebdcab7a5d00d53bbbe46cce45c14bbcfda93f0c/src/logger.ts)
+
+## Supported formats
+
+| Format                                     | Input (array) | Input (stream) | Output (JSON) | Output (text) |
+| ------------------------------------------ | ------------- | -------------- | ------------- | ------------- |
+| JSON                                       | ❌            | ❌             | ✔️            | ✔️            |
+| JSONEachRow                                | ✔️            | ✔️             | ✔️            | ✔️            |
```
mshustov, line 141: Nit: Maybe we can add links to the format description on the company website? https://clickhouse.com/docs/en/interfaces/formats/#jsoneachrow or to a high-level section https://clickhouse.com/docs/en/interfaces/formats

README.md
```diff
@@ -1,8 +1,368 @@
...
+```
+
+### Selecting the data as a stream
+
+```ts
+const rows = await client.query({
+  query: `SELECT * from ${tableName}`,
+  format: 'JSONCompactEachRow',
+})
+for await (const row of rows.asStream()) {
```
mshustov: Now I'm thinking about whether we should rename `asStream()` to `stream()`. WDYT?

README.md
```diff
@@ -1,8 +1,368 @@
...
+  format: 'CSV',
+  query_params: {
+    val1: 10,
+    val2: 20,
+  },
+})
+const result = await rows.text()
+// result is '30\n'
+```
+
```
mshustov, line 585: Let's add an example of specifying settings for a single `query` call.

README.md
```diff
@@ -1,8 +1,368 @@
...
+## Introduction
 
-- Node.JS 14.x or 16.x
+Official Node.js (14.x or 16.x) client for [ClickHouse](https://clickhouse.com/), written purely in TypeScript, thoroughly tested with ClickHouse 22.8 and upcoming 22.9.
+
+It is focused on data streaming for both inserts and selects using standard [Node.js Streaming API](https://nodejs.org/api/stream.html).
+
+As of now, HTTP(s) protocol is used under the hood, with [planned](https://github.com/ClickHouse/clickhouse-js/issues/25) Native protocol support.
+
+## Connection
```
mshustov, line 31: We will need to add `Installation instructions` once we publish the npm package

README.md
```diff
@@ -1,8 +1,368 @@
...
+### Decimal\* types caveats:
+
+Since we do not use `VALUES` clause and there is no additional type conversion, it is not possible to insert Decimal* type columns as strings, only as numbers. This is a suboptimal approach as it might end in float precision loss. Thus, it is recommended to avoid JSON* formats when using Decimals as of now. Consider TabSeparated* / CSV* / CustomSeparated\* formats families for that kind of workflows. Please refer to the [data types tests](https://github.com/ClickHouse/clickhouse-js/blob/c1b70c82f525c39edb3ca1ee05cb5e6b43dba5b3/__tests__/integration/data_types.test.ts#L98-L131) for more concrete examples on how to avoid precision loss.
+
+## Usage examples
+
+### Insert with array input (JSON\* family formats only)
+
+```ts
+await client.insert({
```
mshustov, line 418: Maybe we should describe the client's public interface before examples? For example, we don't show how to create a table before `insert` operation.

README.md
```diff
@@ -1,8 +1,368 @@
...
+
+### Date* / DateTime* types caveats:
+
+Since we use data streaming for inserts without the `VALUES` clause (which does additional type conversion), Date\* type columns can be only inserted as strings and not as Unix time epoch. It can be possibly changed with the future ClickHouse database releases. Please refer to the corresponding [integration tests](https://github.com/ClickHouse/clickhouse-js/blob/ba387d7f4ce375a60982ac2d99cb47391cf76cec/__tests__/integration/date_time.test.ts) for more examples.
+
+### Decimal\* types caveats:
+
+Since we do not use `VALUES` clause and there is no additional type conversion, it is not possible to insert Decimal* type columns as strings, only as numbers. This is a suboptimal approach as it might end in float precision loss. Thus, it is recommended to avoid JSON* formats when using Decimals as of now. Consider TabSeparated* / CSV* / CustomSeparated\* formats families for that kind of workflows. Please refer to the [data types tests](https://github.com/ClickHouse/clickhouse-js/blob/c1b70c82f525c39edb3ca1ee05cb5e6b43dba5b3/__tests__/integration/data_types.test.ts#L98-L131) for more concrete examples on how to avoid precision loss.
+
```
mshustov, line 202: Let's add caveats for Number precision loss in JSON format and call it out ClickHouse returns 64-bit or bigger integers as strings. It can be changed with https://clickhouse.com/docs/en/operations/settings/settings/#json-formats-settings

README.md
```diff
@@ -1,8 +1,368 @@
...
+```
+
+### Selecting the data as JSON including response metadata
+
+```ts
+const rows = await client.query({
+  query: 'SELECT number FROM system.numbers LIMIT 2',
+  format: 'JSON',
+})
+const result = await rows.json<ResponseJSON<{ number: string }>>()
```
mshustov, line 528: Should we provide `Rows` interface details and limitations (JSON is not stream-able)?

README.md
slvrtrn: yes, as a follow-up today.

README.md
slvrtrn: ouch.

README.md
slvrtrn: sounds good

README.md
slvrtrn: Let's remove the mention for now.

README.md
slvrtrn: I'd suggest then:

```ts
// defaults:
// for HTTP protocol, Node.js Agent has infinite max open sockets in the pool
// (when/if we will have Native)
// for Native, we use X max open sockets in the pool
```

README.md
slvrtrn: Let's add the 18.x in the matrix as a follow-up.

PR #90, opened 2022-09-14
Add `ClickHouseClientConfigOptions.max_open_connections` setting

slvrtrn:


__tests__/integration/config.test.ts
```diff
@@ -84,6 +63,61 @@ describe('config', () => {
...
+
+    it('should use several connections', async () => {
+      client = createTestClient({
+        max_open_connections: 2,
+      })
+      void select('SELECT 1 AS x, sleep(0.2)')
+      void select('SELECT 2 AS x, sleep(0.2)')
+      void select('SELECT 3 AS x, sleep(0.2)')
+      void select('SELECT 4 AS x, sleep(0.2)')
+      await retryOnFailure(async () => {
```
mshustov, line 108: optional nit: it might cause flakiness if `'SELECT 2 AS x, sleep(0.2)'` response takes > `'SELECT 3 AS x, sleep(0.2)'` due to a network delay

slvrtrn: Yep, that was my concern as well. Let's keep an eye on it. If it will ever fail, let's find another way to test `maxSockets`.

PR #98, opened 2022-09-21
Move the examples out of README

slvrtrn:
- [x] Extract the examples
- [x] More examples (CSV file streaming, separate select streaming example) 
- [x] Examples README


examples/schema/simple_schema.ts
```diff
@@ -2,7 +2,8 @@ import * as ch from '../../src/schema'
 import type { Infer } from '../../src/schema'
 import { InsertStream } from '../../src/schema'
 import { createClient } from '../../src'
-
+// If you found this example,
+// consider it as a highly experimental WIP development :)
```
mshustov, line 6: 😄

examples/custom_clickhouse_settings.ts
```diff
@@ -0,0 +1,14 @@
+import { createClient } from '@clickhouse/client'
+void (async () => {
+  const client = createClient()
+  const rows = await client.query({
+    query: 'SELECT number FROM system.numbers LIMIT 2',
+    format: 'JSONEachRow',
+    clickhouse_settings: {
+      // See ClickHouseSettings typings
+      connect_timeout: 30,
```
mshustov: nit: it's not a custom setting, but a standard clickhouse setting

examples/endless_flowing_stream_json.ts
```diff
@@ -0,0 +1,78 @@
...
+    process.exit(0)
+  }
+  process.on('SIGINT', cleanup)
+  process.on('SIGTERM', cleanup)
+})()
+
+function pushData(stream: Stream.Readable) {
+  return async () => {
+    console.info('Pushing several records into the stream...')
+    return [...Array(3)].forEach(() => {
```
mshustov, line 71: Optional. Maybe we should insert a larger data set since ClickHouse prefers large inserts?
```js
let batchSize = 1000;
while(batchSize--) {
  stream.push({...})
}
```

examples/insert_file_stream_ndjson.ts
```diff
@@ -0,0 +1,49 @@
...
+  const filename = Path.resolve(
+    process.cwd(),
+    './examples/resources/data.ndjson'
+  )
+
+  await client.insert({
+    table: tableName,
+    values: Fs.createReadStream(filename).pipe(
+      // for JSON* formats, it is currently required
+      // will be unnecessary when "insert" accepts a stream of strings/bytes
```
mshustov: The comment is outdated. `insert` accepts "raw" data, but we want to test `insert` of a stream of objects.

examples/parameter_binding.ts
```diff
@@ -0,0 +1,15 @@
+import { createClient } from '@clickhouse/client'
```
mshustov: Optional: does it make sense to rename the file to `query_with_parameter_binding`?

examples/create_table_cloud.ts
```diff
@@ -0,0 +1,27 @@
+import { createClient } from '@clickhouse/client'
+
+// ClickHouse cloud - a specific CREATE TABLE syntax is allowed
+void (async () => {
+  const client = createClient({
+    host: process.env['CLICKHOUSE_HOST'],
```
mshustov: Let's add a check the host and password aren't empty. The client doesn't validate the config and falls back to the default, which won't work with Cloud.

examples/create_table_cloud.ts
```diff
@@ -0,0 +1,27 @@
+import { createClient } from '@clickhouse/client'
+
+// ClickHouse cloud - a specific CREATE TABLE syntax is allowed
```
mshustov: What is the purpose of `a specific CREATE TABLE syntax is allowed` comment?

mshustov (approved):

examples/create_table_cloud.ts
slvrtrn: Removed

PR #102, opened 2022-09-23
Coverage report and badge

### LINKED ISSUE #42: Collect coverage automatically
* Collect coverage from Jest
* Set certain thresholds (likely 80+%)
* Display a badge in the Readme

slvrtrn:
Resolves #42 

* To save execution time in the pipeline, `integration-tests-local-single-node` now also runs unit tests (practically free) and uploads a coverage report artifact.
* The last job in the pipeline uses the artifact and utilizes the `make-coverage-badge` package. Then, it pushes the generated stuff back to the branch. I limited it to `main` only, as it only makes sense there. Commit message will look like `[skip ci] Update coverage report`.
* Added a badge to README. We will have a link in README attached to the badge that will lead to the HTML preview of the report. An example for this branch: http://htmlpreview.github.io/?https://github.com/ClickHouse/clickhouse-js/blob/coverage-report/coverage/lcov-report/index.html
* Raise the coverage to 90% for style points :)


.github/workflows/run-tests.yml
```diff
@@ -82,9 +82,20 @@ jobs:
         run: |
           npm install
 
+      # Will also run unit tests, but that's almost free.
+      # Otherwise, we need to set up a separate job,
+      # which will also run the integration tests for the second time,
+      # and that's more time-consuming.
       - name: Run integration tests
```
mshustov:
```suggestion
      - name: Run tests
```

.github/workflows/run-tests.yml
```diff
@@ -82,9 +82,20 @@ jobs:
...
         run: |
-          npm run test:integration
+          npm t -- --coverage
+
+      - name: Upload coverage report
+        uses: actions/upload-artifact@v3
+        with:
+          name: coverage
+          path: coverage
+          retention-days: 1
```
mshustov, line 98: Why 1 day?

.github/workflows/run-tests.yml
```diff
@@ -145,3 +156,35 @@ jobs:
...
+          node-version: 16
+      - name: Download coverage report
+        uses: actions/download-artifact@v3
+        with:
+          name: coverage
+          path: coverage
+      - name: Install packages
+        run: npm i -G make-coverage-badge
+      - name: Generate badge
+        run: npx make-coverage-badge
```
mshustov, line 184: Does `npx` require `install` step in advance?

.github/workflows/run-tests.yml
```diff
@@ -82,9 +82,20 @@ jobs:
...
           npm install
 
+      # Will also run unit tests, but that's almost free.
+      # Otherwise, we need to set up a separate job,
+      # which will also run the integration tests for the second time,
+      # and that's more time-consuming.
       - name: Run integration tests
         run: |
-          npm run test:integration
+          npm t -- --coverage
```
mshustov, line 91: Let's add an instruction to `CONTRIBUTING.md` about collecting the coverage locally.

mshustov (approved):

.github/workflows/run-tests.yml
slvrtrn: Oh yeah, I see:

```
Executes <command> either from a local node_modules/.bin, or from a central cache, 
installing any packages needed in order for <command> to run.
```

thanks

slvrtrn: ah, unfortunately it prompts and there seems to be no `--yes` option:

```
➜  ~ npx make-coverage-badge
Need to install the following packages:
  make-coverage-badge@1.2.0
Ok to proceed? (y)
```

.github/workflows/run-tests.yml
slvrtrn: From the doc:

> By default, the artifacts and log files generated by workflows are retained for 90 days before they are automatically deleted. You can adjust the retention period, depending on the type of repository:

> For public repositories: you can change this retention period to anywhere between 1 day or 90 days.

and we need it for maybe 5 minutes. I couldn't find an option to retain it only during the workflow unless we go for a completely custom solution which I prefer to avoid...

PR #104, opened 2022-09-27
add "ping" example and typings for method params

mshustov:
While working on documentation, I noticed a few gaps in examples and typings. More might follow in the following PRs.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -49,7 +49,7 @@ export interface BaseParams {
   clickhouse_settings?: ClickHouseSettings
   /** Parameters for query binding. https://clickhouse.com/docs/en/interfaces/http/#cli-queries-with-parameters */
   query_params?: Record<string, unknown>
-  /** AbortSignal instance to cancel a request in progress. */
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -11,47 +11,66 @@ import { Rows } from './rows'
...
     LoggerClass?: new (enabled: boolean) => Logger
   }
 }
 
 export interface BaseParams {
+  /** ClickHouse settings that can be applied on query level. */
   clickhouse_settings?: ClickHouseSettings
+  /** Parameters for query binding. https://clickhouse.com/docs/en/interfaces/http/#cli-queries-with-parameters */
   query_params?: Record<string, unknown>
+  /** AbortSignal instance to cancel a request in progress. */
```
slvrtrn:
```suggestion
  /** AbortSignal instance (using `node-abort-controller` package) to cancel a request in progress. */
```

mshustov: Ok, added. Note that any compatible implementation does the trick.

PR #105, opened 2022-09-27
use 22.9 for local testing

mshustov:


mshustov:
@slvrtrn do you see any blockers for the current change? We don't know how much time it may take to unblock #100

slvrtrn:
@mshustov no blockers.

PR #108, opened 2022-10-02
Remove split2, refactor Rows for better performance

slvrtrn:
After checking out https://github.com/go-faster/ch-bench and writing a simple test for our client as such:
```ts
import { createClient } from '@clickhouse/client'
;(async () => {
  const client = createClient()
  const rows = await client.query({
    query: 'SELECT number FROM system.numbers_mt LIMIT 500000000',
    format: 'TabSeparated',
  })
  const start = +new Date()
  for await (const _ of rows.stream()) {
    //
  }
  const end = +new Date()
  console.info(`Execution time: ${end - start} ms`)
})()
```

I ended up with an "amazing" performance of around ~280 seconds for executing this bit of code (Ryzen 9 5950X).

So, I did the following:

* Replaced `split2` with a simple self-written solution
* Replaced a `Row` class created on every iteration with just a slim interface
* Replaced return type of `stream()` method with `AsyncGenerator<Row, void>`, and now we have a proper `row` type here:

```ts
  for await (const row of rows.stream()) {
    await row.text() // <-- `row` is actually of type `Row` instead of `any`
  }
```

Before: 285 seconds
After: 96 seconds

Still far from perfect (I will investigate more), but that's a start.

### DIFF
--- a/__tests__/integration/abort_request.test.ts
+++ b/__tests__/integration/abort_request.test.ts
@@ -63,7 +63,7 @@ describe('abort request', () => {
         .then(async (rows) => {
           const stream = rows.stream()
           for await (const chunk of stream) {
-            const [[number]] = chunk.json()
+            const [[number]] = chunk.json<[[string]]>()
             // abort when reach number 3
             if (number === '3') {
               controller.abort()
@@ -86,10 +86,12 @@ describe('abort request', () => {
         .then(async function (rows) {
           const stream = rows.stream()
           for await (const chunk of stream) {
-            const [[number]] = chunk.json()
+            const [[number]] = chunk.json<[[string]]>()
             // abort when reach number 3
             if (number === '3') {
-              stream.destroy()
+              await stream.throw(new Error('boo')).then(() => {
+                //
+              })
             }
           }
         })
--- a/__tests__/integration/abort_request.test.ts
+++ b/__tests__/integration/abort_request.test.ts
@@ -89,9 +89,7 @@ describe('abort request', () => {
             const [[number]] = chunk.json<[[string]]>()
             // abort when reach number 3
             if (number === '3') {
-              await stream.throw(new Error('boo')).then(() => {
-                //
-              })
+              await stream.return()
             }
           }
         })
--- a/__tests__/integration/abort_request.test.ts
+++ b/__tests__/integration/abort_request.test.ts
@@ -93,14 +93,7 @@ describe('abort request', () => {
             }
           }
         })
-      // There was a breaking change in Node.js 18.x behavior
-      if (process.version.startsWith('v18')) {
-        await expect(selectPromise).rejects.toMatchObject({
-          message: 'Premature close',
-        })
-      } else {
-        expect(await selectPromise).toEqual(undefined)
-      }
+      expect(await selectPromise).toEqual(undefined)
     })
 
     // FIXME: it does not work with ClickHouse Cloud.
--- a/__tests__/integration/select.test.ts
+++ b/__tests__/integration/select.test.ts
@@ -50,20 +50,14 @@ describe('select', () => {
   })
 
   describe('consume the response only once', () => {
-    async function assertAlreadyConsumed$<T>(fn: () => Promise<T>) {
+    async function assertAlreadyConsumed<T>(fn: () => Promise<T>) {
       await expect(fn()).rejects.toMatchObject(
         expect.objectContaining({
           message: 'Stream has been already consumed',
         })
       )
     }
-    function assertAlreadyConsumed<T>(fn: () => T) {
-      expect(fn).toThrow(
-        expect.objectContaining({
-          message: 'Stream has been already consumed',
-        })
-      )
-    }
+
     it('should consume a JSON response only once', async () => {
       const rows = await client.query({
         query: 'SELECT * FROM system.numbers LIMIT 1',
@@ -71,9 +65,13 @@ describe('select', () => {
       })
       expect(await rows.json()).toEqual([{ number: '0' }])
       // wrap in a func to avoid changing inner "this"
-      await assertAlreadyConsumed$(() => rows.json())
-      await assertAlreadyConsumed$(() => rows.text())
-      await assertAlreadyConsumed(() => rows.stream())
+      await assertAlreadyConsumed(() => rows.json())
+      await assertAlreadyConsumed(() => rows.text())
+      await assertAlreadyConsumed(async () => {
+        for await (const r of rows.stream()) {
+          r.text()
+        }
+      })
     })
 
     it('should consume a text response only once', async () => {
@@ -83,9 +81,13 @@ describe('select', () => {
       })
       expect(await rows.text()).toEqual('0\n')
       // wrap in a func to avoid changing inner "this"
-      await assertAlreadyConsumed$(() => rows.json())
-      await assertAlreadyConsumed$(() => rows.text())
-      await assertAlreadyConsumed(() => rows.stream())
+      await assertAlreadyConsumed(() => rows.json())
+      await assertAlreadyConsumed(() => rows.text())
+      await assertAlreadyConsumed(async () => {
+        for await (const r of rows.stream()) {
+          r.text()
+        }
+      })
     })
 
     it('should consume a stream response only once', async () => {
@@ -99,9 +101,13 @@ describe('select', () => {
       }
       expect(result).toEqual('0')
       // wrap in a func to avoid changing inner "this"
-      await assertAlreadyConsumed$(() => rows.json())
-      await assertAlreadyConsumed$(() => rows.text())
-      await assertAlreadyConsumed(() => rows.stream())
+      await assertAlreadyConsumed(() => rows.json())
+      await assertAlreadyConsumed(() => rows.text())
+      await assertAlreadyConsumed(async () => {
+        for await (const r of rows.stream()) {
+          r.text()
+        }
+      })
     })
   })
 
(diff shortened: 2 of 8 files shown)


__tests__/integration/select.test.ts
```diff
@@ -328,35 +337,14 @@ describe('select', () => {
...
-        last = chunk.text()
-        i++
-        if (i % 1000 === 0) {
-          stream.pause()
-          setTimeout(() => stream.resume(), 100)
-        }
-      }
-      expect(last).toBe('9999')
-    })
-
```
slvrtrn, line 359: Looks like we don't need this anymore. It can be paused (or not consumed) just on the application level.

__tests__/integration/abort_request.test.ts
```diff
@@ -86,21 +86,14 @@ describe('abort request', () => {
...
           }
         })
-      // There was a breaking change in Node.js 18.x behavior
-      if (process.version.startsWith('v18')) {
-        await expect(selectPromise).rejects.toMatchObject({
-          message: 'Premature close',
-        })
-      } else {
-        expect(await selectPromise).toEqual(undefined)
-      }
```
slvrtrn, line 103: Surprisingly enough, this is no longer the case for Node.js 18.x.

¯\\\_(ツ)_/¯

__tests__/integration/select.test.ts
slvrtrn, line 359: Probably not required anymore with async iterators. It can be paused (not consumed) on the application level, as it is a lazy evaluation.

__tests__/integration/streaming_e2e.test.ts
```diff
@@ -32,13 +31,27 @@ describe('streaming e2e', () => {
...
+              return
+            } else {
+              const json = JSON.parse(line)
+              this.push(json)
+            }
+          })
+        callback()
+      },
+      objectMode: true,
+    })
```
slvrtrn, line 51: Not the best implementation for this (haven't tested how it behaves with large files), but at least it works without `split2`.

__tests__/integration/streaming_e2e.test.ts
```diff
@@ -32,13 +31,27 @@ describe('streaming e2e', () => {
...
+    const jsonTransform = new Transform({
+      transform(data: Buffer, encoding, callback) {
+        data
+          .toString(encoding)
+          .split('\n')
+          .forEach((line) => {
+            if (!line.length) {
+              return
+            } else {
+              const json = JSON.parse(line)
```
slvrtrn, line 44: I think we should support already serialized data somehow. It looks strange to parse it here and then stringify it again in the library.

slvrtrn:
After some digging in Node.js issues I found [this](https://github.com/nodejs/node/issues/31979), and indeed, `for await const of` is about 30% slower than `.on('data', callback)` approach on my machine. 

Using current release version, with the old stream implementation with `split2`, this code

```ts
import { createClient } from '@clickhouse/client';
(async () => {
  const client = createClient()
  const rows = await client.query({
    query: 'SELECT number FROM system.numbers_mt LIMIT 50000000',
    format: 'TabSeparated'
  })
  const start = +new Date()
  const stream = rows.stream()
  stream.on('data', (_) => {
    //
  })
  await new Promise((resolve) => {
    stream.on('end', () => {resolve()})
  })
  const end = +new Date()
  console.info(`Execution time: ${end - start} ms`)
  await client.close()
})()
```
executes in 20 seconds, while this 

```ts
import { createClient } from '@clickhouse/client';
(async () => {
  const client = createClient()
  const rows = await client.query({
    query: 'SELECT number FROM system.numbers_mt LIMIT 50000000',
    format: 'TabSeparated'
  })
  const start = +new Date()
  for await (const _ of rows.stream()) {
    //
  }
  const end = +new Date()
  console.info(`Execution time: ${end - start} ms`)
  await client.close()
})()
```
takes 30 seconds to finish.

So it might be beneficial to return Stream.Readable instead of an AsyncGenerator.
Only need to figure out how to make it faster :)

slvrtrn:
So, with just the allocations removed, keeping the Stream in place
```ts
return Stream.pipeline(
  this._stream,
  split((row: string) => ({ // <- no `new Row` here
    text: row, // <- this is not a function anymore
    json<T>() {
      return decode(row, 'JSON')
    },
  })),
  function pipelineCb(err) {
    if (err) {
      console.error(err)
    }
  }
)
```
using this code
```ts
import { createClient } from '../src'
void (async () => {
  const client = createClient({
    compression: {
      request: false,
      response: false,
    },
  })
  const rows = await client.query({
    query: 'SELECT number FROM system.numbers_mt LIMIT 50000000',
    format: 'TabSeparated',
  })
  const start = +new Date()
  const stream = rows.stream()
  stream.on('data', (_) => {
    //
  })
  await new Promise((resolve) => {
    stream.on('end', () => {
      resolve(0)
    })
  })
  const end = +new Date()
  console.info(`Execution time: ${end - start} ms`)
})()
```
we can get as fast as ~3.5-4 seconds on my machine.
The current release takes ~18 seconds to execute the same code.

On 500M records instead of 50M, that's ~37-38 seconds vs ~280-300 seconds.

src/rows.ts
```diff
@@ -39,62 +38,62 @@ export class Rows {
...
     if (this._stream.readableEnded) {
       throw Error(streamAlreadyConsumedMessage)
     }
-
     validateStreamFormat(this.format)
+    const textDecoder = new TextDecoder()
+    let decodedChunk = ''
+    for await (const chunk of this._stream) {
+      decodedChunk += textDecoder.decode(chunk, { stream: true })
+      let idx = 0
```
mshustov, line 57: nit it's not changed. let's move lower:
```js
const idx = decodedChunk.indexOf('\n')
```

src/rows.ts
```diff
@@ -39,62 +38,62 @@ export class Rows {
...
+             * The method will throw if called on a response in JSON incompatible format.
+             *
+             * It is safe to call this method multiple times.
+             */
+            json<T>(): T {
+              return decode(line, 'JSON')
+            },
+          }
+        } else {
+          break
```
mshustov, line 82: Let's add a special case at the top to remove a nesting level. 
```js
if (idx === -1) break;
const line = decodedChunk.slice(0, idx);
...
```

src/rows.ts
```diff
@@ -39,62 +38,62 @@ export class Rows {
...
+      let idx = 0
+      while (true) {
+        idx = decodedChunk.indexOf('\n')
+        if (idx !== -1) {
+          const line = decodedChunk.slice(0, idx)
+          decodedChunk = decodedChunk.slice(idx + 1)
+          yield {
+            /**
+             * Returns a string representation of a row.
+             */
```
mshustov, line 66: These comments belong to `Row` interface declared below. https://github.com/ClickHouse/clickhouse-js/pull/108/files#diff-b13826a37e4f93783b49eaca9c60dc1d124ee9d6b331be22244f41cc7bb09d39R94-R96
IDE doesn't hint at the method signature.

src/rows.ts
```diff
@@ -39,62 +38,62 @@ export class Rows {
...
+              return decode(line, 'JSON')
+            },
+          }
+        } else {
+          break
         }
       }
-    )
+    }
+    textDecoder.decode() // flush
```
mshustov, line 86: Shouldn't it be consumed? The method typings shows it as
```js
    /**
     * Returns the result of running encoding's decoder. The method can be invoked zero or more times with options's stream set to true, and then once without options's stream (or set to false), to process a fragmented input. If the invocation without options's stream (or set to false) has no input, it's clearest to omit both arguments.
     *
     * ```
     * var string = "", decoder = new TextDecoder(encoding), buffer;
     * while(buffer = next_chunk()) {
     *   string += decoder.decode(buffer, {stream:true});
     * }
     * string += decoder.decode(); // end-of-queue
     * ```
     *
     * If the error mode is "fatal" and encoding's decoder returns error, throws a TypeError.
     */
    decode(input?: BufferSource, options?: TextDecodeOptions): string;
```

__tests__/integration/streaming_e2e.test.ts
mshustov: Let's insert data in `CSV` but query in `JSONCompactEachRow`? It doesn't change the logic much but simplifies the example

__tests__/integration/abort_request.test.ts
mshustov: You removed `destroy()`, which caused the error. Maybe that's the reason?

src/rows.ts
```diff
@@ -39,62 +38,62 @@ export class Rows {
...
+      while (true) {
+        idx = decodedChunk.indexOf('\n')
+        if (idx !== -1) {
+          const line = decodedChunk.slice(0, idx)
+          decodedChunk = decodedChunk.slice(idx + 1)
+          yield {
+            /**
+             * Returns a string representation of a row.
+             */
+            text(): string {
```
mshustov, line 67: Are you sure it's more lightweight than a `Row` class instance?
The method emits an object literal with 2 methods vs. a class with 2 methods in a prototype
```js
class Row {
    text(){}
}
(new Row()).hasOwnProperty('text') // false
const objectLiteral = { 
    text(){}
}
objectLiteral.hasOwnProperty('text') // true
```

slvrtrn:
```ts
split((row: string) => new Row(row, 'JSON')),
```
```bash
ts-node --transpile-only --project tsconfig.dev.json examples/many_numbers.ts
Execution time: 1957 ms
```

```ts
split((row: string) => ({
  text: row,
  json<T>() {
    return decode(row, 'JSON')
  }
})),
```

```bash
ts-node --transpile-only --project tsconfig.dev.json examples/many_numbers.ts
Execution time: 394 ms
```

PR #109, opened 2022-10-03
Breaking changes: refactor Rows for better streaming performance

slvrtrn:
- `split2` moved to dev dependencies only (we can rewrite affected tests and examples later)
- use self-written `Transform` to push `Row` instances in bulk
- **(breaking)** `Rows` renamed to `ResultSet` 
- **(breaking)** every stream iteration now provides `Row[]` instead of `Row`, just as it is returned from CH 
- fixed tests
- fixed docs
- added more examples


__tests__/unit/result.test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    const result: unknown[] = []
+    for await (const rows of rs.stream()) {
+      rows.forEach((row: Row) => {
+        result.push(row.json())
+      })
+    }
+
+    expect(result).toEqual(expectedJson)
+    // expect(stream.readableEnded).toBeTruthy()
```
slvrtrn: that's a bit strange, but it does not have `readableEnded` set to `true` anymore. I wonder if we can fix it.

src/result.ts
```diff
@@ -56,45 +58,57 @@ export class Rows {
...
+            const text = decodedChunk.slice(0, idx)
+            decodedChunk = decodedChunk.slice(idx + 1)
+            rows.push({
+              text,
+              json<T>(): T {
+                return decode(text, 'JSON')
+              },
+            })
+          } else {
+            this.push(rows)
```
mshustov: it will emit an empty array if a batch doesn't contain `\n`

src/result.ts
```diff
@@ -39,14 +36,19 @@ export class Rows {
...
+   *
+   * Should be called only once.
    *
    * The method will throw if called on a response in non-streamable format,
    * and if the underlying stream was already consumed
-   * by calling the other methods
+   * by calling the other methods.
    */
-  stream(): Stream.Readable {
+  stream(): Readable {
```
mshustov: It used to be `Stream.Readable` to avoid confusion with global [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/ReadableStream)

src/result.ts
```diff
@@ -56,45 +58,57 @@ export class Rows {
...
 
-    return Stream.pipeline(
-      this._stream,
-      // only JSON-based format are supported at the moment
-      split((row: string) => new Row(row, 'JSON')),
-      function pipelineCb(err) {
-        if (err) {
-          console.error(err)
+    let decodedChunk = ''
+    const toRows = new Transform({
```
mshustov, line 65: Don't we need `flush` logic anymore? https://github.com/mcollina/split2/blob/master/index.js#L61

src/result.ts
```diff
@@ -56,45 +58,57 @@ export class Rows {
...
+          }
         }
-      }
-    )
+        callback()
+      },
+      objectMode: true,
+    })
+
+    return this._stream.pipe(toRows, { end: true })
```
mshustov: nodejs docs recommend `Stream.pipeline` usage https://nodejs.org/api/stream.html#streampipelinesource-transforms-destination-callback over `stream.pipe`:
> A module method to pipe between streams and generators forwarding errors and properly cleaning up and provide a callback when the pipeline is complete.

src/result.ts
slvrtrn: As I understand, it is guaranteed to have `\n` termination for every row CH returns. Let's leave it as is for now.

examples/select_streaming_on_data.ts
```diff
@@ -0,0 +1,32 @@
+import type { Row } from '../src'
+import { createClient } from '../src'
+
+/**
+ * Can be used for consuming large datasets for reducing memory overhead,
+ * or if your response exceeds built in Node.js limitations,
+ * such as 512Mb for strings
```
mshustov: Let's add a link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length
> In V8 (used by Chrome and Node), the maximum length is 229 - 24 (~1GB). On 32-bit systems, the maximum length is 228 - 16 (~512MB).

mshustov (approved):

examples/select_streaming_on_data.ts
slvrtrn: Ouch. After I added the link, I noticed that the info is probably outdated there. The limit on Node.js for 64-bit systems is 512M now, down from 1Gb in the past.

Issue #26, opened 2022-08-07
Extend logging support


mshustov:
- make logging level configurable
- add logs to improve request traceability
    - connection opened / closed
    - request timed out
    - request aborted
    - retry request

slvrtrn:
I'd say this is resolved; the latest change was in #237, which included even more improvements in the logging mechanism.

PR #110, opened 2022-10-04
Better logging support

slvrtrn:
As I was debugging failing integration tests vs Cloud, I added more logs on response and made `Logger` more flexible ("structured" args and error logging).

Additionally, logs are muted for successful tests now and printed for failing ones, including the test name for every log entry to ease the debugging.

I  should probably add an example of using it with [pino](https://github.com/pinojs/pino) or another logger as a follow-up with full #26 proposed features coverage.

Sample log from the tests (with just `console.debug`)

```
2022-09-30T17:26:20.1729568Z     console.debug
2022-09-30T17:26:20.1730026Z       [insert stream (JSON formats) JSONCompactEachRow should work with JSONCompactStringsEachRowWithNames]
2022-09-30T17:26:20.1730489Z       [HTTP Adapter] Got a response from ClickHouse {
2022-09-30T17:26:20.1730904Z         request_method: 'POST',
2022-09-30T17:26:20.1731195Z         request_path: '/',
2022-09-30T17:26:20.1731867Z         request_params: '?insert_quorum=2&database_replicated_enforce_synchronous_settings=1&wait_end_of_query=1&database=clickhousejs__ef86ab70382443cb8790b2a6b6df98b6',
2022-09-30T17:26:20.1732554Z         request_headers: {},
2022-09-30T17:26:20.1734086Z         response_status: 200,
2022-09-30T17:26:20.1734320Z         response_headers: {
2022-09-30T17:26:20.1734672Z           date: 'Fri, 30 Sep 2022 17:26:05 GMT',
2022-09-30T17:26:20.1735006Z           connection: 'Keep-Alive',
2022-09-30T17:26:20.1735384Z           'content-type': 'text/plain; charset=UTF-8',
2022-09-30T17:26:20.1736071Z           'x-clickhouse-server-display-name': 'clickhouse-cloud',
2022-09-30T17:26:20.1736481Z           'transfer-encoding': 'chunked',
2022-09-30T17:26:20.1736820Z           'keep-alive': 'timeout=3',
2022-09-30T17:26:20.1737467Z           'x-clickhouse-summary': '{"read_rows":"3","read_bytes":"189","written_rows":"0","written_bytes":"0","total_rows_to_read":"3","result_rows":"0","result_bytes":"0"}'
2022-09-30T17:26:20.1737822Z         },
2022-09-30T17:26:20.1738048Z         response_time_ms: 1105
2022-09-30T17:26:20.1738268Z       }
```

Related to #26


src/connection/adapter/base_http_adapter.ts
```diff
@@ -294,17 +294,30 @@ export abstract class BaseHttpAdapter implements Connection {
...
 
   private logResponse(
+    request: Http.ClientRequest,
     params: RequestParams,
     response: Http.IncomingMessage,
     startTimestamp: number
   ) {
+    const getRequestHeaders = () => {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const { authorization, ...headers } = request.getHeaders()
```
mshustov: There might be other sensitive headers. Maybe we should create an allow-list of headers to log?

src/connection/adapter/base_http_adapter.ts
```diff
@@ -294,17 +294,30 @@ export abstract class BaseHttpAdapter implements Connection {
...
+      return headers
+    }
+    const getRequestBody = () => {
+      return typeof params.body === 'string' ? params.body : 'Stream'
```
mshustov: I'm not sure it's safe to log the body as it might contain sensitive data to insert into ClickHouse. Let's remove it.

src/connection/adapter/base_http_adapter.ts
```diff
@@ -294,17 +294,30 @@ export abstract class BaseHttpAdapter implements Connection {
   }
 
   private logResponse(
+    request: Http.ClientRequest,
     params: RequestParams,
     response: Http.IncomingMessage,
     startTimestamp: number
   ) {
+    const getRequestHeaders = () => {
```
mshustov: why not inline values instead of creating functions on every call? It adds GC pressure

slvrtrn: Ouch, that's the case indeed. I thought it will be at least hoisted. Turned out it is not correct for this definition.

src/connection/adapter/base_http_adapter.ts
slvrtrn: I'd suggest maybe adding a possibility to enable or disable it in the logger, along with the fields whitelist, cause it might be useful for debugging purposes.

src/connection/adapter/base_http_adapter.ts
slvrtrn: yes, we should indeed

src/connection/adapter/base_http_adapter.ts
mshustov: > I'd suggest maybe adding a possibility to enable or disable it in the logger, along with the fields whitelist, cause it might be useful for debugging purposes.

I'd avoid extending public API. We can add an allow-list option for `body` logging on a request.

__tests__/integration/config.test.ts
```diff
@@ -5,12 +5,16 @@ import type { RetryOnFailureOptions } from '../utils/retry'
...
+    message: string
+    err?: Error
+    args?: Record<string, unknown>
+  }[] = []
 
   afterEach(async () => {
     await client.close()
-    if (messages.length) {
-      messages = []
+    if (logs.length) {
```
mshustov: Why not reset `logs` unconditionally?

__tests__/integration/config.test.ts
```diff
@@ -49,7 +53,13 @@ describe('config', () => {
       },
     })
     await client.ping()
-    expect(messages).toContainEqual(expect.stringContaining('GET /ping'))
+    expect(logs[0]).toEqual({
```
mshustov: nit: we need a test for every log level

src/logger.ts
```diff
@@ -1,23 +1,23 @@
 export class Logger {
   constructor(readonly enabled = false) {}
```
mshustov: Should we add support for `log level` to log only messages with a level > configured?

src/connection/adapter/base_http_adapter.ts
```diff
@@ -294,17 +294,30 @@ export abstract class BaseHttpAdapter implements Connection {
...
     const duration = Date.now() - startTimestamp
-
-    this.logger.debug(
-      `[http adapter] response: ${params.method} ${params.url.pathname}${
-        params.url.search ? ` ${params.url.search}` : ''
-      } ${response.statusCode} ${duration}ms`
-    )
+    this.logger.debug('[HTTP Adapter] Got a response from ClickHouse', {
+      request_method: params.method,
+      request_path: params.url.pathname,
```
mshustov: We can keep both, but in the case of CH, request_method and request_path are almost always `POST` and `/` ( for `ping` method they are `GET` and `/ping`)

mshustov (approved):

src/logger.ts
```diff
@@ -1,23 +1,113 @@
...
+    if (this.logLevel < (ClickHouseLogLevel.OFF as number)) {
+      this.logger.info(params)
+    }
   }
 
-  error(message: string): void {
-    if (!this.enabled) return
-    console.error(message)
+  private getClickHouseLogLevel(): ClickHouseLogLevel {
+    const fromEnv = process.env['CLICKHOUSE_LOG_LEVEL']
```
mshustov:
```suggestion
    const logLevel = process.env['CLICKHOUSE_LOG_LEVEL']
```

src/logger.ts
```diff
@@ -1,23 +1,113 @@
...
+    if (this.logLevel < (ClickHouseLogLevel.INFO as number)) {
+      this.logger.debug(params)
+    }
+  }
 
-  debug(message: string): void {
-    if (!this.enabled) return
-    console.log(message)
+  info(params: LogParams): void {
+    if (this.logLevel < (ClickHouseLogLevel.WARN as number)) {
```
mshustov: why not
```js
 if (this.logLevel <= (ClickHouseLogLevel.INFO as number)) {
```

IMO it's easy to read as `info` matches `INFO` level

src/logger.ts
```diff
@@ -1,23 +1,113 @@
...
+      this.logger.info(params)
+    }
   }
 
-  info(message: string): void {
-    if (!this.enabled) return
-    console.log(message)
+  warn(params: LogParams): void {
+    if (this.logLevel < (ClickHouseLogLevel.ERROR as number)) {
+      this.logger.info(params)
```
mshustov: shouldn't it be `this.logger.warn(params)`? the same about `error` method

src/client.ts
```diff
@@ -119,7 +120,7 @@ function normalizeConfig(
     clickhouse_settings: config.clickhouse_settings ?? {},
     log: {
       enable: loggingEnabled,
```
mshustov: Should we remove the flag?

__tests__/integration/config.test.ts
```diff
@@ -41,26 +44,48 @@ describe('config', () => {
...
+        message: 'Got a response from ClickHouse',
+        args: expect.objectContaining({
+          request_path: '/ping',
+          request_method: 'GET',
+        }),
+      })
+    })
+
+    it('should provide a custom logger implementation (but logs are disabled)', async () => {
+      process.env[logLevelKey] = 'OFF'
```
mshustov, line 107: could you add a test for every log level please?
also let's verify the `console.log` renders an error message and points to an error stack trace

PR #114, opened 2022-10-06
Rework `enable_http_compression` behaviour in the client

slvrtrn:
* Do not enforce `enable_http_compression=1` setting if `compression.response` set to `true` - it is now required to have it either in the server's `users.xml` or supply it separately
* Update the README


slvrtrn:
Please ignore.

PR #118, opened 2022-10-18
TLS support

### LINKED ISSUE #52: SSL support
Basic HTTPS support was implemented in #22 
As a follow-up, we need to add support for https://clickhouse.com/docs/en/guides/sre/ssl-user-auth/

slvrtrn:
Resolves #52 

OpenSSL certificates support. See https://clickhouse.com/docs/en/guides/sre/configuring-ssl/
Test certificates are taken from https://github.com/ClickHouse/clickhouse-connect/tree/main/tests/tls
Currently, CI tests are running on a single node only.

Tested with local setup and Yandex cloud.


genzgd (approved):
You could probably trim down the users.xml and config.xml files to remove default or unneeded values just to allow focus on the important settings.  (logger, quotes, etc.0

slvrtrn:
@genzgd sure, that's indeed a good one for a follow-up.

PR #122, opened 2022-10-25
improve Typescript compatibility

KMahoney:
These small changes make the library compatible with much earlier versions of Typescript.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,13 +1,10 @@
 import Stream from 'stream'
-import type { TLSParams } from './connection'
-import { type Connection, createConnection } from './connection'
+import type { Connection, TLSParams } from './connection'
+import { createConnection } from './connection'
 import { Logger } from './logger'
 import { isStream, mapStream } from './utils'
-import {
-  type DataFormat,
-  encodeJSON,
-  isSupportedRawFormat,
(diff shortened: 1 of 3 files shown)


slvrtrn:
I'd be happy to merge this if this is required for your use case but I have a few concerns:
- The main one is: how do we verify that the imports are organized in a proper way for future commits if we use only modern versions of TS for the development and CI? I doubt that we want to add an extra matrix run for that.
- What are the blockers for the users to upgrade their TS version?
- Is it possible to set up WebStorm's "optimize imports" feature to comply with the older import syntax?

KMahoney:
There's no need to merge on my behalf. This just solved an immediate problem for me, and I thought it might open up the library to more users.

We (@causalhq) probably will upgrade Typescript soon anyway. Tomorrow I'll check if it's as simple as bumping the number :)

slvrtrn:
@KMahoney yep, please let me know.
If it helps immediately, I will just merge this one and release a new tag with your changeset.
So there will be at least one version that works with the older TS versions.
But it will not be guaranteed that the imports will not be "broken" for the old versions of TS in the future.

slvrtrn:
Closing this due to inactivity.

Issue #130, opened 2023-01-25
Query metrics or query id


ultram4rine:
Hi everyone! I'm trying to implement showing query metrics for ultram4rine/sqltools-clickhouse-driver#281 and since I'm rewriting my driver using this official lib I have a question, is it possible to get `query_id` to use it in statement like:
```sql
SELECT query_duration_ms,
  read_rows,
  read_bytes
FROM system.query_log
WHERE query_id = 'abfba2b6-9488-4c45-9186-d86b4c0d8bba'
  AND type = 2;
```
If no, is it possible to implement getting query id or metrics in this lib?

slvrtrn:
if you are using JSON format for queries, it should be possible to use it like this: 

```bash

**Comment by slvrtrn** [comment]
@ultram4rine I wonder if something like this will help? 

`query_id` will be obtainable from `query`, `insert` and `exec`

https://github.com/ClickHouse/clickhouse-js/blob/ada4dae87f4a7335a4509dc13c5a38eb11e2886a/__tests__/integration/query_log.test.ts#L13-L60

**Comment by ultram4rine** [comment]
@slvrtrn while `query_id` is pretty useful, using `system.query_log` is a bit awkward. On the other hand, with `JSON` format, I get statistics just like in console or in web interface + meta columns, but I can't use stream.
I wonder, if it is possible to add this statistics and meta to stream, like it was in [@alpa/node-clickhouse](https://github.com/apla/node-clickhouse#returns-duplexstream) lib?

**Comment by ultram4rine** [comment]
@slvrtrn yes, but maybe stats like
```js
stream.on('end', () => {
  console.log(
    stream.statistics.elapsed_time,
    stream.statistics.read_rows,         stream.statistics.read_bytes,
  )
})
```

slvrtrn:
@ultram4rine I'll check if it's possible with HTTP protocol and streams

slvrtrn:
@ultram4rine our best bet here is to use `X-ClickHouse-Progress` header (see the entry about it in https://clickhouse.com/docs/en/interfaces/http#default-database) plus add some internal stopwatch to mimic the `elapsed_time` behavior as it is not supported there;
however, I am not sure how useful it will be.

mshustov:
> r best bet here is to decode the X-ClickHouse-Progress header (see the entry about it in https://clickhouse.com/docs/en/interfaces/http#default-database)

[...]

PR #132, opened 2023-01-30
Add random query_id to each query

slvrtrn:
## Summary
A PoC for https://github.com/ClickHouse/clickhouse-js/issues/130

## Checklist

clickhouse-docs will be updated once merged


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 0.0.12
+
+### New features
+* `ResultSet.query_id` contains a unique query identifier that might be useful for retrieving query metrics from `system.query_log`
+
+### Breaking changes
+* `client.insert` now returns `string` (query_id) instead of `void`
```
mshustov: Why not `{ query_id: string }` to make it extendable in the future?

__tests__/unit/result.test.ts
```diff
@@ -1,6 +1,7 @@
 import type { Row } from '../../src'
 import { ResultSet } from '../../src'
 import Stream, { Readable } from 'stream'
+import { guid } from '../utils'
```
mshustov, line 4: Could you remind why we generate `query_id` instead of reading from `X-ClickHouse-Query-Id` header?

CHANGELOG.md
slvrtrn: that's great

__tests__/unit/result.test.ts
slvrtrn: right, it will be even simpler this way

slvrtrn: it is actually a bit more complicated than that.
in my tests, it seems that the header is returned only for SELECT statements;
for anything else, we need to generate our own.
I prefer to keep the generated one for every method for consistency

mshustov (approved):
let's update CHANGELOG and merge the PR

PR #133, opened 2023-01-30
Set the User-Agent header according to the language client spec

slvrtrn:
## Summary
Set the User-Agent header according to the [language client spec](https://docs.google.com/document/d/1924Dvy79KXIhfqKpi1EBVY3133pIdoMwgCQtZ-uhEKs/edit#heading=h.ah33hoz5xei2)

## Checklist
- [ ] Update the changelog when #132 is merged to avoid the conflicts

### DIFF
--- a/__tests__/unit/http_adapter.test.ts
+++ b/__tests__/unit/http_adapter.test.ts
@@ -8,6 +8,7 @@ import { HttpAdapter } from '../../src/connection/adapter'
 import { retryOnFailure, TestLogger } from '../utils'
 import { getAsText } from '../../src/utils'
 import { LogWriter } from '../../src/logger'
+import { BaseHttpAdapter } from '../../src/connection/adapter/base_http_adapter'
 
 describe('HttpAdapter', () => {
   const gzip = Util.promisify(Zlib.gzip)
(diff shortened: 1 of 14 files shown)


src/version.ts
```diff
@@ -0,0 +1 @@
+export default '0.0.11'
```
mshustov, line 1: What if we read the version directly from package.json? Is it okay to have 2 sources of the value synchronised automatically?

__tests__/unit/http_adapter.test.ts
```diff
@@ -239,6 +240,42 @@ describe('HttpAdapter', () => {
...
+      }
+      public getDefaultHeaders() {
+        return this.headers
+      }
+    }
+
+    it('should have proper user agent without app id', async () => {
+      const myHttpAdapter = new MyTestHttpAdapter()
+      const headers = myHttpAdapter.getDefaultHeaders()
+      expect(headers['Authorization']).toMatch(/^Basic [A-Za-z0-9/+=]+$/)
```
mshustov: What is the purpose of this assertion in `User-Agent` test?

__tests__/unit/http_adapter.test.ts
```diff
@@ -239,6 +240,42 @@ describe('HttpAdapter', () => {
...
+      expect(headers['User-Agent']).toMatch(
+        /^clickhouse-js\/[0-9\\.]+? \(lv:nodejs\/v[0-9\\.]+?; os:(?:linux|darwin|win32)\)$/
```
mshustov, line 248: optional: I'd remove the assertion against `linux|darwin|win32` to prevent flakiness. We can match a non-empty string instead.

mshustov (approved):

__tests__/unit/http_adapter.test.ts
slvrtrn: let's move it to a separate test case then

src/version.ts
slvrtrn: Initially, I wanted to read the package.json to keep it simple. However, to do that, you need to read that file (`const packageJson = require('./package.json')` for example) , and this library is intended to be used within other apps with their own package.json file; I prefer to avoid possible bugs due to the resolution of the relative path.

__tests__/unit/http_adapter.test.ts
slvrtrn: I doubt that it will be flaky.
https://nodejs.org/api/os.html#os_os_platform
> Possible values are 'aix', 'darwin', 'freebsd','linux', 'openbsd', 'sunos', and 'win32'.

That's very unlikely that the code will be run on `aix`, `openbsd`, `freebsd` or `sunos`

PR #139, opened 2023-03-13
Allow query_id override, remove unnecessary examples

slvrtrn:
## Summary
* `query_id` can be now overridden for all main client's methods: `query`, `exec`, `insert`.
* Superfluous code examples were removed

## Checklist


examples/stream_created_from_array_json.ts
```diff
@@ -1,38 +0,0 @@
-import { createClient } from '@clickhouse/client'
```
slvrtrn, line 1: I think in this case stream is pretty much useless and just an array could be used (and we have these examples already)

examples/insert_json_stream.ts
```diff
@@ -1,44 +0,0 @@
-import { createClient } from '@clickhouse/client'
```
slvrtrn, line 1: "endless" stream examples have broader coverage

examples/stream_created_from_array_raw.ts
```diff
@@ -1,32 +0,0 @@
-import { createClient } from '@clickhouse/client'
```
slvrtrn, line 1: same as with `stream_created_from_array_json.ts`

__tests__/integration/select.test.ts
```diff
@@ -321,17 +332,6 @@ describe('select', () => {
...
-      it.skip('returns queryId in response', async () => {
-        const rs = await client.query({
-          query: 'SELECT number FROM system.numbers LIMIT 5',
-          format: 'JSON',
-        })
-
-        const response = await rs.json<ResponseJSON<{ number: string }>>()
-
-        expect(response.query_id).toBeInstanceOf(Number)
-      })
```
slvrtrn, line 334: This is an outdated version of the same test above

examples/stream_created_from_array_raw.ts
mshustov: That's the only example of inserting serialized data. I'd keep it

mshustov (approved):

PR #160, opened 2023-05-16
Fix NULL parameters binding

slvrtrn:
## Summary
The issue was reported some time ago in the community Slack. NULL has to be passed as an actual null symbol (`\N`) instead of a string `'NULL'`. [See a similar fix in clickhouse-connect](https://github.com/ClickHouse/clickhouse-connect/pull/166/files#diff-b324bcc9f3a437c5fd9d073c6d2f04d74a4b598fe631594410f6abe9e0993c04R450).

## Checklist

### DIFF
--- a/__tests__/integration/select_query_binding.test.ts
+++ b/__tests__/integration/select_query_binding.test.ts
@@ -1,3 +1,4 @@
+import type { QueryParams } from '../../src'
 import { type ClickHouseClient } from '../../src'
 import { createTestClient } from '../utils'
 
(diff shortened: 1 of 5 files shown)


__tests__/integration/select_query_binding.test.ts
```diff
@@ -268,4 +269,35 @@ describe('select with query binding', () => {
...
+        query_params: {
+          n: null,
+        },
+      })
+
+      const response = await rs.text()
+      expect(response).toBe('0\n1\n2\n')
+    })
+
+    it('should with an explicit undefined', async () => {
```
mshustov, line 291: How does it behave with an undefined value?

mshustov (approved):

__tests__/integration/select_query_binding.test.ts
slvrtrn: Missing key will fail and there is a test for that, cause the parameter is not bound at all. Explicit undefined and null values are sent as `\\N`

Issue #162, opened 2023-05-23
Remove platform-specific dependencies


maxprilutskiy:
**Issue Description:**

I would like to raise a concern regarding the current state of the `@clickhouse/client` package. 

It heavily relies on platform-specific dependencies, including `http`, `zlib`, and `stream`, which limits its usability in certain environments, particularly serverless edge setups. I believe it is crucial to address this issue and propose a solution that would make `@clickhouse/client` more versatile and compatible with a broader range of platforms.

**Problem Statement:**

The presence of node-specific dependencies in `@clickhouse/client` prevents its usage in environments where these dependencies are either unavailable or incompatible. For instance, serverless edge environments require platform-agnostic libraries to ensure seamless execution across different platforms. Unfortunately, the current state of `@clickhouse/client` hinders its potential in such scenarios.

**Desired Outcome:**

I kindly request the development team to reconsider the dependencies in `@clickhouse/client` and strive for a solution that eliminates any platform-specific ties. By doing so, we can make the client package more adaptable, enabling it to be used in a wider array of environments, including serverless edge deployments. This change would empower developers to leverage the benefits of the ClickHouse client in scenarios where these platform-specific dependencies are not available or compatible.

**Proposed Solution:**

To address this issue effectively, I propose refactoring `@clickhouse/client` to remove direct dependencies on node-specific packages like `http`, `zlib`, and `stream`. Instead, I recommend employing platform-agnostic alternatives or providing a pluggable interface that allows users to supply their own implementations for the underlying communication layer. Or, `fetch` API could just be used instead, that is supported everywhere nowadays.

**Benefits:**

- Enhanced compatibility: Eliminating platform-specific dependencies will broaden the compatibility of `@clickhouse/client`, making it viable for a wider range of platforms and environments.
- Serverless edge deployment: Making the client compatible with serverless edge environments empowers users to harness the advantages of ClickHouse for real-time analytics and data processing in these setups.
- Flexibility: Introducing a pluggable interface for the underlying communication layer will allow users to customize and adapt the client to meet their specific requirements.

**Additional Context:**

Serverless edge computing has gained significant popularity due to its scalability and reduced infrastructure management overhead. By enabling the usage of `@clickhouse/client` in these environments, we can provide developers with a powerful tool for processing and analyzing data close to the edge, ultimately improving performance and reducing latency.

[...]

Issue #153, opened 2023-04-27
Support for environments without access to `http`


chitalian:
### Describe the bug
I am using cloudflare workers and within their environment there is no support to `http`.

When we import we get this error:

### Steps to reproduce
1. wrangler init <name>
2. cd <name>
3. yarn add @clickhouse/client
4. add this line `import { createClient as clickhouseCreateClient } from "@clickhouse/client";`
5. wrangler dev --local

### Expected behaviour

We should migrate to `fetch` or create a new client that uses `fetch` instead of `http`

### Error log

```bash
[mf:err] TypeError: globalThis.XMLHttpRequest is not a constructor
    at checkTypeSupport (/Users/justin/Documents/repos/promptzero/helicone/worker/node_modules/rollup-plugin-node-polyfills/polyfills/http-lib/capability.js:20:11)
    at node_modules/rollup-plugin-node-polyfills/polyfills/http-lib/capability.js (/Users/justin/Documents/repos/promptzero/helicone/worker/node_modules/rollup-plugin-node-polyfills/polyfills/http-lib/capability.js:39:45)
    at __init (/private/var/folders/zc/g8ym72nn3tl70yg68f5khk040000gn/T/tmp-10247-HoWOjJp9q0Bf/index.js:8:56)
    at node_modules/rollup-plugin-node-polyfills/polyfills/http-lib/request.js (/Users/justin/Documents/repos/promptzero/helicone/worker/node_modules/rollup-plugin-node-polyfills/polyfills/http-lib/request.js:1:1)
    at __init (/private/var/folders/zc/g8ym72nn3tl70yg68f5khk040000gn/T/tmp-10247-HoWOjJp9q0Bf/index.js:8:56)
    at node-modules-polyfills:http (/var/folders/zc/g8ym72nn3tl70yg68f5khk040000gn/T/tmp-10247-HoWOjJp9q0Bf/node-modules-polyfills:http:30:1)
    at __init (/private/var/folders/zc/g8ym72nn3tl70yg68f5khk040000gn/T/tmp-10247-HoWOjJp9q0Bf/index.js:8:56)
    at node-modules-polyfills-commonjs:http (/var/folders/zc/g8ym72nn3tl70yg68f5khk040000gn/T/tmp-10247-HoWOjJp9q0Bf/node-modules-polyfills-commonjs:http:2:18)
[...]

PR #163, opened 2023-05-29
WIP: Refactor client to support custom connections

slvrtrn:
## Summary

Refactor the client to support custom connections. 

NB: I haven't implemented connections other than the existing Node.js one yet; the first attempt (browser) is planned as a follow-up. This is just the necessary restructuring before we can make some progress there.

Related issues: 
* https://github.com/ClickHouse/clickhouse-js/issues/162 
* https://github.com/ClickHouse/clickhouse-js/issues/153

The code now is a small monorepo, divided into several workspaces
* common (platform-independent types, data/query formatting, and other utils)
* client (the main client with all the config parsing and validation, without the connection layer, depends on the common workspace)
* node connection (the actual Node.js implementation, which depends on the client workspace)

The idea is that we can create new connections (fetch/axios/whatever) as just a new workspace and run CI using all of them one by one (run all tests with node connections, run a subset of tests for the browser, etc).

This also means we will have multiple NPM packages - the base client itself without a connection implementation included (replacing current `@clickhouse/client`) and at least one package per connection (`@clickhouse/client-node`, `@clickhouse/client-fetch`). The end user will have to install only one, though. This PR does not address the release routine (need to address the imports/paths, adjust GitHub Actions, etc)

Once it is released (when at least one new connection is made), the changes for the existing Node.js users should be minimal, just a simple change of import (`createClient` will come from `@clickhouse/client-node` instead of `@clickhouse/client`).

For example, with `@clickhouse/client-node`:

```ts
import { createClient } from '@clickhouse/client-node'
const client = createClient({
  host: getFromEnv('CLICKHOUSE_HOST'),
  password: getFromEnv('CLICKHOUSE_PASSWORD'),
})
```

See the [workspace source](https://github.com/ClickHouse/clickhouse-js/blob/c5166c367e78916c4c53a47e327d54cc9ba95dc2/packages/node_connection/src/client.ts#L19-L35) for more details.

However, if just `@clickhouse/client` is added as a dependency, it is now possible to implement a custom version of the `Connection` interface and pass it with client configuration options. 

For example:

```ts
const client = new ClickHouseClient({
  host: 'http://<host>:<port>',
  connection: (config) => new MyCustomConnection(config),
})
```

Where `MyCustomConnection` implements

```ts
export interface Connection {
  ping(): Promise<boolean>
  close(): Promise<void>
  query(params: BaseQueryParams): Promise<QueryResult>
  exec(params: BaseQueryParams): Promise<QueryResult>
  insert(params: InsertParams): Promise<InsertResult>
}
```

See [connection.ts](https://github.com/ClickHouse/clickhouse-js/blob/c5166c367e78916c4c53a47e327d54cc9ba95dc2/packages/common/src/connection.ts#L36-L64) for more details.

Currently, we have `stream` and `Buffer` as leftover Node.js-specific types in the common module. However, there are polyfills available:
 
* Streams: https://www.npmjs.com/package/web-streams-polyfill 
* Buffer: https://www.npmjs.com/package/buffer

Again, as a follow-up, I will investigate what is possible for the browser or, for example, Cloudflare workers since these are the common use cases.

## Checklist


slvrtrn:
Changed the base branch from `main` to `0.1.0` cause I expect that it will take several PRs before we have it ready.

slvrtrn:
Closing this in favor of #165

Issue #17, opened 2022-08-04
Improve testing infrastructure


mshustov:
- [x] collocate test files with the code
- [x] improve assertions in async tests. done in https://github.com/ClickHouse/clickhouse-js/issues/30

slvrtrn:
No further improvements required after #165

PR #165, opened 2023-06-08
Generic client, Browser + Node.js connections

### LINKED ISSUE #2: Provide a version for browser environment
The browser version of the client can provide limited functionality but with a similar API.
The integration tests must be run in both `nodejs` and `browser` envs.

In the first implementation, the client uses `webpack` to shim `http` module with a browser-compatible version. This approach doesn't allow us to implement env-specific logic for the network modules, so let's consider implementing a separate `http_browser_adapter` that might encapsulate browser-specific network logic. The separation can be done on the code level by using a dedicated `browser` entry point of `package.json` see https://docs.npmjs.com/cli/v8/configuring-npm/package-json#browser

slvrtrn:
## Summary

Further efforts after #163 (different approach).

Introduce browser connection (using native fetch) with no Node.js modules in the common interfaces. No polyfills are required. It will also be possible to write new connections on top of `@clickhouse/client-common` if necessary.

The client was refactored into three packages:
- `@clickhouse/client-common`: all possible platform-independent code, types and interfaces
- `@clickhouse/client-browser`: new "browser" (or non-Node.js env) connection, uses native fetch. Depends on the common package.
- `@clickhouse/client`: Node.js connection as it was before. Depends on the common package.

Related issues: 
* https://github.com/ClickHouse/clickhouse-js/issues/162 
* https://github.com/ClickHouse/clickhouse-js/issues/153

Additionally, resolves #2, #17 

Jest was replaced with jasmine + jasmine-expect, cause it's easier to maintain the tests that way - Karma does not like Jest very much; however, one of few downsides of that approach is that `retryOnFailure` does not work anymore as jasmine matchers do not throw an exception, thus in some tests I had to use janky hardcoded `sleep` calls.

Current browser connection limitations: 
- Browser connection insert streams won't be supported atm because it does not work across all the platforms (see [FF](https://bugzilla.mozilla.org/show_bug.cgi?id=1387483) issue); arrays work fine though; select streams are also fine;
- Custom Logger currently is not implemented for the browser connection;
- Custom user-agent (with application name) is not implemented for browser connection (need to investigate what's actually possible there. Currently, I don't set it at all).

## Checklist


README.md
```diff
@@ -6,10 +6,6 @@
 <p align="center">
 <a href="https://github.com/ClickHouse/clickhouse-js/actions/workflows/tests.yml">
 <img src="https://github.com/ClickHouse/clickhouse-js/actions/workflows/tests.yml/badge.svg?branch=main">
-</a>
-<a href="http://htmlpreview.github.io/?https://github.com/ClickHouse/clickhouse-js/blob/main/coverage/lcov-report/index.html">
-<img src="./coverage/badge.svg">
-</a>
```
mshustov, line 12: I think we need to update README to explain the purpose of all 3 packages. Right now it says
> <h1 align="center">ClickHouse Node.JS client</h1>

webpack.config.js
```diff
@@ -0,0 +1,63 @@
...
+      '.js', // for 3rd party modules in node_modules
+    ],
+    plugins: [
+      new TsconfigPathsPlugin({
+        configFile: 'tsconfig.dev.json',
+        logLevel: 'ERROR',
+      }),
+    ],
+    fallback: {
+      buffer: false,
```
mshustov: Do we even get any of these in the browser bundle? How? I'd rather throw an exception to make sure `nodejs` close doesn't leak into the browser bundle

tsconfig.dev.json
```diff
@@ -11,9 +11,14 @@
...
     "noUnusedParameters": false,
     "outDir": "dist",
-    "baseUrl": ".",
+    "baseUrl": "./",
     "paths": {
-      "@clickhouse/client": ["./src/index.ts"]
+      "@clickhouse/client-common": ["packages/client-common/src/index.ts"],
+      "@clickhouse/client-common/*": ["packages/client-common/src/*"],
+      "@clickhouse/client": ["packages/client-node/src/index.ts"],
+      "@clickhouse/client/*": ["packages/client-node/src/*"],
```
mshustov: Can we avoid deep imports? Since it extends API surface, I'd consider 
- exporting public API only from the top level
- use relative paths to import inside a module

packages/client-common/src/logger.ts
```diff
@@ -63,7 +63,10 @@ export class LogWriter {
   }
 
   private getClickHouseLogLevel(): ClickHouseLogLevel {
-    const logLevelFromEnv = process.env['CLICKHOUSE_LOG_LEVEL']
+    const logLevelFromEnv =
+      typeof process !== 'undefined'
```
mshustov: maybe we should extract it into `isBrowser` predicate for better self-documentation?

package.json
```diff
@@ -23,26 +23,27 @@
...
-    "test:integration:cloud": "CLICKHOUSE_TEST_ENVIRONMENT=cloud jest --runInBand --testPathPattern=__tests__/integration --globalSetup='<rootDir>/__tests__/setup.integration.ts'",
+    "test": "ts-node -r tsconfig-paths/register --project=tsconfig.dev.json node_modules/jasmine/bin/jasmine --config=jasmine.all.json",
+    "test:unit": "ts-node -r tsconfig-paths/register --project=tsconfig.dev.json node_modules/jasmine/bin/jasmine --config=jasmine.unit.json",
+    "test:integration": "ts-node -r tsconfig-paths/register --project=tsconfig.dev.json node_modules/jasmine/bin/jasmine --config=jasmine.integration.json",
+    "test:integration:local_cluster": "CLICKHOUSE_TEST_ENVIRONMENT=local_cluster npm run test:integration",
+    "test:integration:cloud": "CLICKHOUSE_TEST_ENVIRONMENT=cloud npm run test:integration",
+    "test:browser": "karma start karma.config.cjs",
+    "test:browser:integration:local_cluster": "CLICKHOUSE_TEST_ENVIRONMENT=local_cluster npm run test:browser",
+    "test:browser:integration:cloud": "CLICKHOUSE_TEST_ENVIRONMENT=cloud npm run test:browser",
```
mshustov, line 36: Shouldn't we add a command to create a browser build?

package.json
```diff
@@ -52,19 +53,24 @@
     "eslint-config-prettier": "^8.6.0",
     "eslint-plugin-prettier": "^4.2.1",
     "husky": "^8.0.2",
-    "jest": "^29.4.0",
```
mshustov, line 54: We removed `jest`, let's update `CONTRIBUTING.md`. can be done as a follow-up.

package.json
```diff
@@ -52,19 +53,24 @@
...
+    "karma-chrome-launcher": "^3.2.0",
+    "karma-jasmine": "^5.1.0",
+    "karma-sourcemap-loader": "^0.4.0",
+    "karma-typescript": "^5.5.4",
+    "karma-webpack": "^5.0.0",
     "lint-staged": "^13.1.0",
     "prettier": "2.8.3",
     "split2": "^4.1.0",
-    "ts-jest": "^29.0.5",
+    "ts-jest": "^29.1.0",
```
mshustov, line 63: Shouldn't we remove it if we don't use `jest`? Or is it necessary for `jasmin`?

karma.config.cjs
```diff
@@ -0,0 +1,53 @@
...
+    basePath: '',
+    frameworks: ['webpack', 'jasmine'],
+    // list of files / patterns to load in the browser
+    files: [
+      'karma.setup.cjs',
+      '__tests__/integration/*.test.ts',
+      '__tests__/integration/browser/*.test.ts',
+      '__tests__/utils/*.ts',
+      '__tests__/unit/*.test.ts',
+      '__tests__/unit/browser/*.test.ts',
```
mshustov: Is there a reason not to keep platform-specific tests in an appropriate package?

examples/select_streaming_on_data.ts
```diff
@@ -1,5 +1,5 @@
-import type { Row } from '../src'
-import { createClient } from '../src'
+import { createClient } from '@clickhouse/client'
+import type { Row } from '@clickhouse/client-common'
```
mshustov: What if we re-export some of the `common` interfaces? Asking users to install 2 packages is not the best DX

coverage/coverage-summary.json
```diff
@@ -1,35 +0,0 @@
-{"total": {"lines":{"total":597,"covered":553,"skipped":0,"pct":92.62},"statements":{"total":639,"covered":590,"skipped":0,"pct":92.33},"functions":{"total":188,"covered":166,"skipped":0,"pct":88.29},"branches":{"total":293,"covered":253,"skipped":0,"pct":86.34},"branchesTrue":{"total":0,"covered":0,"skipped":0,"pct":100}}
```
mshustov: Are we going to re-implement test coverage later? Could you create a follow-up issue?

__tests__/utils/test_logger.ts
```diff
@@ -23,16 +26,17 @@ function formatMessage({
...
-  try {
-    return expect.getState().currentTestName || 'Unknown'
-  } catch (e) {
-    // ReferenceError can happen here cause `expect`
-    // is not yet available during globalSetup phase,
-    // and we are not allowed to import it explicitly
-    return 'Global Setup'
-  }
-}
+// function getTestName() {
```
mshustov: can be removed?

__tests__/utils/retry.ts
```diff
@@ -39,9 +39,9 @@ export async function retryOnFailure<T>(
   return await attempt()
 }
 
-function sleep(ms: number): Promise<void> {
+export function sleep(ms: number): Promise<void> {
   return new Promise((resolve) => {
-    setTimeout(resolve, ms).unref()
```
mshustov, line 44: Doesn't it prevent nodejs process from exiting?

karma.setup.cjs
```diff
@@ -0,0 +1,6 @@
+window.it.skip = (name) => {
```
mshustov: Do we use either of them?

packages/client-browser/src/client.ts
```diff
@@ -0,0 +1,22 @@
...
+): ClickHouseClient<ReadableStream> {
+  return new ClickHouseClient<ReadableStream>({
+    makeConnection: (params: ConnectionParams) => new BrowserConnection(params),
+    makeResultSet: (
+      stream: ReadableStream,
+      format: DataFormat,
+      query_id: string
+    ) => new ResultSet(stream, format, query_id),
+    valuesEncoder: new BrowserValuesEncoder(),
+    ...(config || {}),
```
mshustov, line 45:
```suggestion
    ...(config ?? {}),
```

packages/client-browser/src/connection/browser_connection.ts
```diff
@@ -0,0 +1,178 @@
...
+      params,
+      searchParams,
+    })
+    return {
+      query_id,
+    }
+  }
+
+  async ping(): Promise<boolean> {
+    // TODO: catch an error and just log it, returning false?
```
mshustov, line 98: maybe it's not a bad idea, considering that request might fail due to CORS, for example

packages/client-browser/src/connection/browser_connection.ts
```diff
@@ -0,0 +1,178 @@
...
+        keepalive: !bodyIsStream,
+        method: method ?? 'POST',
+        signal: abortController.signal,
+        headers: withCompressionHeaders({
+          headers: this.defaultHeaders,
+          compress_request: false,
+          decompress_response: this.params.compression.decompress_response,
+        }),
+        // @ts-expect-error 'duplex' does not exist in type 'RequestInit'
+        duplex: bodyIsStream ? 'half' : undefined, // https://developer.chrome.com/articles/fetch-streaming-requests/
```
mshustov: Does it work on firefox? Does it even work with chrome? No, according to the comment https://github.com/ClickHouse/clickhouse-js/pull/165/files#diff-2c9af14af89d1014d64398505017be35255b65ad8a383b0228dee69c9f6eabdeR6-R8
Maybe we shouldn't support `insert` streaming?

karma.config.cjs
```diff
@@ -0,0 +1,53 @@
...
+      '__tests__/integration/browser/*.ts': ['webpack', 'sourcemap'],
+      '__tests__/utils/*.ts': ['webpack', 'sourcemap'],
+    },
+    reporters: ['progress'],
+    port: 9876,
+    colors: true,
+    logLevel: config.LOG_INFO,
+    autoWatch: false,
+    // available browser launchers: https://npmjs.org/browse/keyword/karma-launcher
+    browsers: ['Chrome_without_security'],
```
mshustov, line 45: Just curious: are we going to test in Firefox in the future?

__tests__/integration/browser/browser_exec.test.ts
```diff
@@ -0,0 +1,47 @@
+import type { ClickHouseClient } from '@clickhouse/client-common'
+import { createTestClient } from '../../utils'
+import { getAsText } from '../../../packages/client-browser/src/utils' // FIXME: Karma does not like "proper" typescript path here
```
mshustov: another reason to keep the test in the appropriate package

__tests__/utils/client.ts
```diff
@@ -36,21 +46,39 @@ export function createTestClient(
...
-      database,
+    const localConfig: BaseClickHouseClientConfigOptions<Stream> = {
+      database: databaseName,
       ...logging,
       ...config,
       clickhouse_settings: clickHouseSettings,
-    })
+    }
+    if (process.env.browser) {
+      // @ts-ignore
```
mshustov: Optional since it's a test file but let's use `@ts-expect-error`

__tests__/integration/data_types.test.ts
```diff
@@ -82,53 +80,53 @@ describe('data types', () => {
...
-    })
-    const result = await client
-      .query({
-        query: `SELECT * FROM ${table} ORDER BY id ASC`,
-        format: 'TabSeparated',
-      })
-      .then((r) => r.text())
-    expect(result).toEqual(row1 + row2)
-  })
+  // it('should work with decimals', async () => {
```
mshustov: Any actions for this test

__tests__/integration/config.test.ts
```diff
@@ -44,147 +42,91 @@ describe('config', () => {
...
-      expect(logs).toHaveLength(0)
-    })
-  })
-
-  describe('max_open_connections', () => {
-    let results: number[] = []
-    afterEach(() => {
-      results = []
-    })
+  // describe('Logger support', () => {
```
mshustov: What do you think about these commented-out tests?

mshustov (approved):
LGTM overall. What are our next steps?
There are a few follow-ups to address, at least a simple browser test for memory leaks, update docs, release beta versions, and gather feedback. Did I miss anything?

webpack.config.js
slvrtrn: That's a weird behavior of tests - Webpack still wants to assemble the rest of the Node.js code (I think because of how I am creating the test client) and complaints about incompatibilities. These warnings are muted when all polyfills are explicitly disabled - it seems it finally just skips the code now.

tsconfig.dev.json
slvrtrn: > exporting public API only from the top level

it is already like that, isn't it?

> use relative paths to import inside a module

it is like that within a particular connector (`import ... from '../utils`), but common imports are done as if it was in a separate library (cause that's exactly how it will be when it's released)

package.json
slvrtrn: It is necessary for `jasmine`, unfortunately.

karma.config.cjs
slvrtrn: Can be moved later; decided to keep it simple for now.

__tests__/utils/retry.ts
slvrtrn: it does; thanks for pointing this out - I removed this when I was trying to check how it works with jasmine (it does not). Will bring this back.

karma.setup.cjs
slvrtrn: Nope, will remove

packages/client-browser/src/client.ts
slvrtrn: Isn't it the same in this case?

packages/client-browser/src/connection/browser_connection.ts
slvrtrn: I think I will just remove streaming support for browser inserts since it does not work in FF _at all_.

The question is whether it's better to make InsertValues based on the connection then. Cause for Node.js it will be Array or `Stream.Readable`, and for the browser it will be just Array.

I will check whether it impacts DX if we make it generic on the client type as well, e.g. instead of 
```ts
class ClickHouseClient<Stream = unknown>
```

it might be 

```ts
type BaseInsertValues = ReadonlyArray<T> | InputJSON<T> | InputJSONObjectEachRow<T>
class ClickHouseClient<Stream = unknown, InsertValues = BaseInsertValues>
```

karma.config.cjs
slvrtrn: We should, I think. There are differences in streaming support. Once this one is close to the release, let's create the issues for leftovers.

__tests__/integration/browser/browser_exec.test.ts
slvrtrn: This is a good one

__tests__/integration/config.test.ts
slvrtrn: Forgot about this one. It should be moved to Node.js-only tests for now.

__tests__/integration/data_types.test.ts
slvrtrn: yes, I need to check this one out, thanks

packages/client-browser/src/connection/browser_connection.ts
slvrtrn: Nested/dependent generics were not trivial to implement for this scenario. I ended up just restricting the type returned by the browser's `createClient` method.

package.json
slvrtrn: Created an issue https://github.com/ClickHouse/clickhouse-js/issues/176

coverage/coverage-summary.json
slvrtrn: Done, https://github.com/ClickHouse/clickhouse-js/issues/177

packages/client-browser/src/connection/browser_connection.ts
slvrtrn: Added a follow-up https://github.com/ClickHouse/clickhouse-js/issues/178

karma.config.cjs
slvrtrn: A follow-up: https://github.com/ClickHouse/clickhouse-js/issues/179

slvrtrn:
@mshustov, I added several follow-ups to reduce the scope of this one (sorted by priority):
- https://github.com/ClickHouse/clickhouse-js/issues/179
- https://github.com/ClickHouse/clickhouse-js/issues/180
- https://github.com/ClickHouse/clickhouse-js/issues/176
- https://github.com/ClickHouse/clickhouse-js/issues/178
- https://github.com/ClickHouse/clickhouse-js/issues/177

CHANGELOG.md
```diff
@@ -1,15 +1,52 @@
+## 0.1.1
```
mshustov, line 1: Do you want to keep the development in `0.2.0` branch until all the packages are stabilized?

README.md
```diff
@@ -1,22 +1,26 @@
...
+Official JS client for [ClickHouse](https://clickhouse.com/), written purely in TypeScript,
+thoroughly tested with actual ClickHouse versions.
+
+The repository consists of three packages:
 
-It is focused on data streaming for both inserts and selects using standard [Node.js Streaming API](https://nodejs.org/docs/latest-v14.x/api/stream.html).
+- `@clickhouse/client` - Node.js client, built on top of [HTTP](https://nodejs.org/api/http.html)
+  and [Stream](https://nodejs.org/api/stream.html) APIs; supports streaming for both selects and inserts.
+- `@clickhouse/client-browser` - browser client, built on top of [Fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)
+  and [Streams](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) APIs; supports streaming for selects.
```
mshustov:
```suggestion
  and [Web Streams](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) APIs; supports streaming for selects.
```

packages/client-browser/__tests__/integration/browser_abort_request.test.ts
```diff
@@ -0,0 +1,76 @@
+import type { ClickHouseClient, Row } from '@clickhouse/client-common'
+import { createTestClient } from '@test/utils'
+
+// FIXME: abort signal stopped working.
+//  To be revisited in https://github.com/ClickHouse/clickhouse-js/issues/177
+xdescribe('Browser abort request streaming', () => {
```
mshustov: Hasn't it been fixed?

packages/client-common/__tests__/utils/client.ts
```diff
@@ -36,21 +46,42 @@ export function createTestClient(
...
-    })
+    }
+    if (process.env.browser) {
+      return require('../../../client-browser/src/client').createClient(
+        cloudConfig
+      )
+    } else {
+      // props to https://stackoverflow.com/a/41063795/4575540
+      // @ts-expect-error
+      return eval('require')('../../../client-node/src/client').createClient(
```
mshustov, line 64: 🙈  let's create a follow-up issue to fix it

mshustov (approved):

CHANGELOG.md
slvrtrn: Yes, CHANGELOG will be there once everything is finalized.

packages/client-browser/__tests__/integration/browser_abort_request.test.ts
slvrtrn: I forgot to push it, thanks

packages/client-common/__tests__/utils/client.ts
slvrtrn: https://github.com/ClickHouse/clickhouse-js/issues/185

PR #167, opened 2023-06-14
0.1.0 - Timeouts rework, command method

### LINKED ISSUE #161: 3 seconds delay when max_open_connections is set
### Describe the bug
When `max_open_connections` value is set, client delays request by 3 seconds.

### Steps to reproduce
1. Create client with `max_open_connections = 1`  ( with 1 it's easier to notice, but I think the problem exists with any value )
2. Run some insert queries in parallel or sequentially
3. Notice that the delay is 3 seconds for each request

There seems to be no problem with selects.

### Expected behaviour
Requests should be executed much faster

### Code example
The code below runs two sequentual queries and logs response times.

```
const { createClient } = require('@clickhouse/client');

(async () => {
    const client = new createClient({
        host: 'http://localhost:8123',
        max_open_connections: 1,
    });

    await client.exec({
        query: 'DROP TABLE IF EXISTS test_table',
    });
    await client.exec({
        query: `CREATE TABLE test_table (user_id UInt64) ENGINE = MergeTree() PRIMARY KEY (user_id) FORMAT TabSeparated`,
    });

    console.log('table created');

    for (let i = 0; i < 2; ++i) {
        const start = performance.now();

        try {
            await client.insert({
                table: 'test_table',
                values: [
                    {
                        user_id: i,
                    }
                ],
                format: 'JSONEachRow',
            });

            console.log('insert success', i);
        } finally {
            const end = performance.now();

            console.log('request took', end - start, 'ms');
        }
    }

    const countResult = await client.query({
        query: 'SELECT count() as count FROM test_table',
    });
    const countJson = await countResult.json();

    console.log('finished, rows after insert:', countJson.data[0].count);
})();
```

### Error log
Notice that each request takes 3 seconds
```
table created
insert success 0
request took 3008.7007999999914 ms
insert success 1
request took 3008.5150999999605 ms
finished, rows after insert: 2
```

#### Environment
* Client version: 0.0.16
* Language version: nodejs v18.12.1
* OS: Ubuntu 22.04

#### ClickHouse server
* ClickHouse Server version: 23.4.2.11-alpine
* ClickHouse Server non-default settings, if any: None
* `CREATE TABLE` statements for tables involved: Included in code example

Here is the docker compose file used to run clickhouse server
```
version: '3.7'

services:
  clickhouse:
    image: "clickhouse/clickhouse-server:23.4.2.11-alpine"
    restart: always
    ports:
      - "8123:8123"
      - "9000:9000"
```

### Additional notes
I have example with axios client with roughly same setup which does not have this problem.

Axios version: 1.4.0

Logs ( notice much faster response time )
```
table created
insert success 0
request took 6.3756999999750406 ms
insert success 1
request took 3.766200000187382 ms
finished, rows after insert: 2
```

Here is the code
```
const http = require('http');

const axios = require('axios');

const axiosInstance = axios.create({
    httpAgent: new http.Agent({
        keepAlive: true,
        maxSockets: 1,
    }),
});

const baseRequestParams = {
    auth: {
        "username": "default",
        "password": "",
    },
    url: 'http://localhost:8123',
    method: 'post',
};

(async () => {
    await axiosInstance({
        ...baseRequestParams,
        "data": "DROP TABLE IF EXISTS test_table FORMAT TabSeparated",
    });

    await axiosInstance({
        ...baseRequestParams,
        "data": "CREATE TABLE test_table (user_id UInt64) ENGINE = MergeTree() PRIMARY KEY (user_id) FORMAT TabSeparated",
    });

    console.log('table created');

    for (let i = 0; i < 2; ++i) {
        const start = performance.now();

        try {
            await axiosInstance({
                ...baseRequestParams,
                "data": `INSERT INTO test_table  (user_id)  VALUES (${i})`,
            });

            console.log('insert success', i);
        } catch (error) {
            console.error('request error', error);
        } finally {
            const end = performance.now();

            console.log('request took', end - start, 'ms');
        }
    }

    const countResult = await axiosInstance({
        ...baseRequestParams,
        "data": "SELECT count() as c FROM test_table",
    });

    console.log('finished, rows after insert:', countResult.data);
})();
```

slvrtrn:
## Summary
* Resolves #161
* `node-abort-controller` dependency is removed
* `insert` response stream is destroyed immediately
* `command` method is introduced
* We are now relying on signals that are passed to the request instead of setting sockets timeouts

## Breaking changes

* `connect_timeout` client setting is removed, as it was unused in the code.
 
## New features

* `command` method is introduced as an alternative to `exec`.
`command` does not expect user to consume the response stream, and it is destroyed immediately.
Essentially, this is a shortcut to `exec` that destroys the stream under the hood.
Consider using `command` instead of `exec` for DDLs and other custom commands which do not provide any valuable output.

Example:

```ts
// incorrect: stream is not consumed and not destroyed, request will be timed out eventually
await client.exec('CREATE TABLE foo (id String) ENGINE Memory')

// correct: stream does not contain any information and just destroyed
const { stream } = await client.exec('CREATE TABLE foo (id String) ENGINE Memory')
stream.destroy()

// correct: same as exec + stream.destroy()
await client.command('CREATE TABLE foo (id String) ENGINE Memory')
```

### Bug fixes

* Fixed delays on subsequent requests after calling `insert` that happened due to unclosed stream instance when using low number of `max_open_connections`. See [#161](https://github.com/ClickHouse/clickhouse-js/issues/161) for more details.

## Checklist
- [x] For significant changes, documentation in https://github.com/ClickHouse/clickhouse-docs was updated with further explanations or tutorials - see https://github.com/ClickHouse/clickhouse-docs/pull/1252


CHANGELOG.md
```diff
@@ -1,3 +1,34 @@
...
+  query: `SELECT 1 FORMAT CSV`,
+  returnResponseStream: false,
+})
+
+// returns { query_id, stream } - pre-0.1.0 behavior
+// Important: if the response stream is requested, the user is expected to consume the stream
+// otherwise, the request will be eventually timed out
+await client.exec({
+  query: `SELECT 1 FORMAT CSV`,
+  returnResponseStream: true,
```
mshustov: What if we consider adding a new method instead? For dev using `*.js` without type-checking, it might be tricky to handle

src/connection/adapter/base_http_adapter.ts
```diff
@@ -107,20 +105,32 @@ export abstract class BaseHttpAdapter implements Connection {
...
       const start = Date.now()
 
-      const request = this.createClientRequest(params.url, params)
-      request.once('socket', (socket) => {
-        socket.setTimeout(this.config.request_timeout)
+      const abortController = params?.abort_controller || new AbortController()
+      let isTimedOut = false
+      const timeout = setTimeout(() => {
+        isTimedOut = true
+        abortController.abort('Request timed out')
```
mshustov: How does `abortController` behaves if external code calls `abort` on the already aborted controller?

src/connection/adapter/https_adapter.ts
```diff
@@ -40,12 +39,13 @@ export class HttpsAdapter extends BaseHttpAdapter implements Connection {
 
   protected createClientRequest(
     url: URL,
-    params: RequestParams
+    params: RequestParams & { abort_controller: AbortController }
```
mshustov: it might be another parameter as well:
```typescript
    url: URL,
    params: RequestParams,
    { abort_controller: AbortController }
```

CHANGELOG.md
slvrtrn: I believe the vast majority of the use cases are DDLs and tricky inserts here, where we do not expect a response stream (and it won't be there by default). And even when we get it, it's just a raw stream without any transformations from `ResultSet`. In the very beginning, I was not planning to expose the `exec` response at all, but I added it there just in case.

src/connection/adapter/base_http_adapter.ts
slvrtrn: Like, timeout and then abort?

src/client.ts
```diff
@@ -16,8 +21,6 @@ import type { InputJSON, InputJSONObjectEachRow } from './clickhouse_types'
 export interface ClickHouseClientConfigOptions {
   /** A ClickHouse instance URL. Default value: `http://localhost:8123`. */
   host?: string
-  /** The timeout to set up a connection in milliseconds. Default value: `10_000`. */
-  connect_timeout?: number
```
slvrtrn, line 20: I will add this to CHANGELOG as well. It is removed because it was actually never used.

src/connection/adapter/base_http_adapter.ts
mshustov: yes, for example. the chance is small though

CHANGELOG.md
slvrtrn: In the end, I implemented a new method as suggested.

__tests__/integration/abort_request.test.ts
```diff
@@ -53,13 +52,59 @@ describe('abort request', () => {
...
+      controller.abort('bar qaz') // no-op, does not throw here
+    })
+
     it('cancels a select query while reading response', async () => {
       const controller = new AbortController()
       const selectPromise = client
         .query({
           query: 'SELECT * from system.numbers',
           format: 'JSONCompactEachRow',
-          abort_signal: controller.signal as AbortSignal,
```
mshustov, line 62: Can we keep `abort_signal` as a part of the external API? I feel like asking to provide `abort_controller` is a leaking implementation detail. We can create an abort controller internally and compose it with the external one (if provided). see https://github.com/whatwg/fetch/issues/905

CHANGELOG.md
```diff
@@ -1,3 +1,34 @@
+## 0.1.0
+
+## Breaking changes
+
+* `connect_timeout` client setting is removed, as it was unused in the code.
```
mshustov, line 5: Is it worth listing as a breaking change? It might fail TS compilation, but shouldn't affect the runtime

src/connection/adapter/base_http_adapter.ts
```diff
@@ -178,37 +181,41 @@ export abstract class BaseHttpAdapter implements Connection {
...
       }
 
-      request.on('response', onResponse)
-      request.on('timeout', onTimeout)
-      request.on('error', onError)
-      request.on('abort', onAbort)
-      request.on('close', onClose)
+      abortController.signal.addEventListener(
+        'abort',
+        () => {
```
mshustov: How do we remove the subscription? Might it cause a memory leak?

slvrtrn: it's fixed already by https://github.com/ClickHouse/clickhouse-js/pull/167/commits/a796f8f6283d645ee367724cf4dee786eb449522

CHANGELOG.md
slvrtrn: Let's keep it there.

mshustov (approved):

Issue #150, opened 2023-04-25
Socket hang up after 2nd query


GGo3:
### Get this error after 2nd query to clickhouse
```
Error: socket hang up
    at connResetException (node:internal/errors:691:14)
    at Socket.socketOnEnd (node:_http_client:471:23)
    at Socket.emit (node:events:402:35)
    at endReadableNT (node:internal/streams/readable:1343:12)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  code: 'ECONNRESET'
}
```

### Steps to reproduce
1. Make 1 query with results
2. Make some actions with stat less than 7-10 sec
3. After that make one more query and get this error

Also, try to increase connect_timeout to 20 sec but it didn't help.

slvrtrn:
What is the environment where the code is executed and the Node.js version?
Could you provide a minimal reproduction scenario?

GGo3:
Sure
Node v16.13.2
@clickhouse/client 0.0.14
Clickhouse server 21.7.4.18

in createClient use only host username password database from .env
The problem is that when we have one open CH pool, we make select queries to the database one after another. When the interval between these queries is not long, up to 1-2 seconds, everything works smoothly. However, as soon as a longer interval is made between the queries (more then 5 sek), we receive this error. It feels like a feature was added where if the connection is not used for some time, the socket cannot be used.

olexiyb:
We also try to understand the cause of ECONNRESET we see randomly
This article gives some [clue](https://connectreport.com/blog/tuning-http-keep-alive-in-node-js/)

slvrtrn:
@GGo3, do you have a load balancer in front of ClickHouse? It could be one of the reasons for timeouts. Additionally, we just released 0.0.15 (props to @olexiyb); you could try checking it out, maybe the fix will work for you as well.

GGo3:
Nope. Just remote clickhouse server, every 10 min by crontab start node script, createClient to CH and work with data.

GGo3:
Updated new version 0.0.15. It doesn't help

slvrtrn:
@GGo3, have you tried increasing `request_timeout`?

slvrtrn:
I tried to reproduce it locally, but it worked fine.

My env: Fedora 37, Node.js v16.17.0, ClickHouse 23.3, @clickhouse/client 0.0.15

```js
import {createClient} from '@clickhouse/client'

void (async () => {
    const client = createClient({
[...]

PR #169, opened 2023-06-27
0.1.1: Keep-Alive sockets housekeeping

slvrtrn:
## Summary

This is a potential fix for #150.

* Introduce a mechanism to detect a potentially expired socket on the client side when using Keep-Alive. If such socket is detected, and retry is enabled in the configuration, both socket and request will be immediately destroyed, and the client will try again.
* Allow disabling Keep-Alive on the client side entirely.
* TRACE log level added

It'd be nice to have a `freeSocketTimeout` setting to remove expired idling sockets in the background before we even try to use them (such as in `agentkeepalive`). Still, even there, it was not working as I expected.

## Checklist
- [x] For significant changes, documentation in https://github.com/ClickHouse/clickhouse-docs was updated with further explanations or tutorials


src/connection/adapter/base_http_adapter.ts
```diff
@@ -83,14 +84,25 @@ function isDecompressionError(result: any): result is { error: Error } {
...
+const expiredSocketMessage = 'expired socket'
+
 export abstract class BaseHttpAdapter implements Connection {
   protected readonly headers: Http.OutgoingHttpHeaders
+  private readonly retry_strategy?: RetryStrategy
+  private readonly known_sockets = new Map<
+    string,
+    {
+      last_used_time: number
+      timeout_id?: NodeJS.Timeout
```
slvrtrn: This is a rudiment, as I was experimenting with background expiring idling sockets cleanup. Will remove.

src/connection/retry_strategy.ts
```diff
@@ -0,0 +1,33 @@
...
+  async waitForRetry(): Promise<void> {
+    return undefined
+  },
+}
+
+/**
+ * Simple strategy to immediately retry a failed request
+ * until we exceed {@link max_retries} number of requests.
+ */
+export class SimpleRetryStrategy implements RetryStrategy {
```
slvrtrn: I will add tests if we keep it.

__tests__/unit/http_adapter.test.ts
```diff
@@ -172,7 +173,8 @@ describe('HttpAdapter', () => {
     })
 
     describe('request compression', () => {
-      it('sends a compressed request if compress_request: true', async () => {
+      // FIXME: this one hangs for an unknown reason
+      it.skip('sends a compressed request if compress_request: true', async () => {
```
slvrtrn: This is the only failing test so far.

__tests__/integration/config.test.ts
```diff
@@ -60,7 +60,7 @@ describe('config', () => {
     })
 
     it('should use the default logger implementation', async () => {
-      process.env[logLevelKey] = 'DEBUG'
+      process.env[logLevelKey] = 'TRACE'
```
mshustov: why is it necessary?

src/connection/adapter/base_http_adapter.ts
```diff
@@ -83,14 +84,25 @@ function isDecompressionError(result: any): result is { error: Error } {
   return result.error !== undefined
 }
 
+const expiredSocketMessage = 'expired socket'
+
 export abstract class BaseHttpAdapter implements Connection {
   protected readonly headers: Http.OutgoingHttpHeaders
+  private readonly retry_strategy?: RetryStrategy
+  private readonly known_sockets = new Map<
```
mshustov: can we use [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) to avoid potential memory leaks?

src/connection/retry_strategy.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+/**
+ * Default strategy for the client. Never retries the requests.
+ */
+export const NoRetryStrategy: RetryStrategy = {
+  shouldRetry() {
+    return false
+  },
+  async waitForRetry(): Promise<void> {
```
mshustov: Is it for the future to support different retry strategies?

src/connection/adapter/base_http_adapter.ts
```diff
@@ -159,11 +195,84 @@ export abstract class BaseHttpAdapter implements Connection {
...
+              config.keep_alive.socket_ttl
+            if (isPossiblyExpired) {
+              logger.trace({
+                module: 'HTTP Adapter',
+                message: 'Socket should be expired - terminate it',
+              })
+              knownSockets.delete(socket.id as string)
+              socket.destroy() // immediately terminate the connection
+              request.destroy()
+              reject(expiredSocketMessage)
```
mshustov: Minor: Even though JS allows to throw any type as an exception, let's follow the [best practices](https://github.com/goldbergyoni/nodebestpractices#-22-extend-the-built-in-error-object) and deal with `Error` object only.

src/connection/adapter/base_http_adapter.ts
```diff
@@ -159,11 +195,84 @@ export abstract class BaseHttpAdapter implements Connection {
...
+              pipeStream()
+            }
+          }
+        } else {
+          const socketId = uuid.v4()
+          logger.trace({
+            module: 'HTTP Adapter',
+            message: `Using a new socket ${socketId}`,
+          })
+          Object.assign(socket, { id: socketId })
```
mshustov: I'd avoid mutating the platform's built-in objects since it might break or be broken at any time. There is a pattern of using WeakMap to attach private metadata to any object. Let's use it. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap#associating_metadata

src/connection/adapter/base_http_adapter.ts
```diff
@@ -159,11 +195,84 @@ export abstract class BaseHttpAdapter implements Connection {
...
+      const knownSockets = this.known_sockets
+      const logger = this.logger
       function onSocket(socket: net.Socket): void {
-        // Force KeepAlive usage (workaround due to Node.js bug)
-        // https://github.com/nodejs/node/issues/47137#issuecomment-1477075229
-        socket.setKeepAlive(true, 1000)
+        // if socket is reused
+        if ('id' in socket) {
+          logger.trace({
+            module: 'HTTP Adapter',
```
mshustov: nit: it's `Base Http Adapter`

src/connection/adapter/base_http_adapter.ts
```diff
@@ -159,11 +195,84 @@ export abstract class BaseHttpAdapter implements Connection {
...
+        }
+
+        if (params.compress_request) {
+          Stream.pipeline(bodyStream, Zlib.createGzip(), request, callback)
+        } else {
+          Stream.pipeline(bodyStream, request, callback)
+        }
+      }
+
+      // avoid `this` confusion in `onSocket` closure
```
mshustov: Is it the case if we declare `onSocket` as an arrow function? 
```typescript
const onSocket = (socket: net.Socket) => ...
```

mshustov (commented):
LGTM

src/client.ts
```diff
@@ -43,6 +52,27 @@ export interface ClickHouseClientConfigOptions {
...
+    /** Enable or disable HTTP Keep-Alive mechanism. Default: true */
+    enabled?: boolean
+    /** Sockets housekeeping settings */
+    sockets?: {
+      /** How long to keep a particular open socket alive (in milliseconds).
+       * Should be less than the server setting
+       * (see `keep_alive_timeout` in server's `config.xml`).
+       * Default value: 2500
+       * (based on the default ClickHouse server setting, which is 3000) */
+      socket_ttl: number
```
mshustov: Why not timeout to follow the standard term? To avoid mixing it up with the built-in socket timeout?

slvrtrn: Yes, the `timeout` term is used a bit differently - that one will happen when there is no _data_ sent to the socket after X milliseconds.

__tests__/unit/http_adapter.test.ts
mshustov: are we going to fix it before the release?

src/client.ts
```diff
@@ -14,35 +14,62 @@ import type { ClickHouseSettings } from './settings'
...
   username?: string
-  /** The user password. Default: ''. */
+  /** The user password. <br/> Default: ''. */
   password?: string
-  /** The name of the application using the nodejs client. Default: empty. */
+  /** The name of the application using the nodejs client.
+   * <br/> Default: empty. */
   application?: string
-  /** Database name to use. Default value: `default`. */
+  /** Database name to use. <br/> Default value: `default`. */
```
mshustov, line 43: Why is it necessary? for a suggestion tooltip in an IDE?

src/connection/adapter/base_http_adapter.ts
```diff
@@ -110,10 +123,31 @@ export abstract class BaseHttpAdapter implements Connection {
...
+  private async request(
+    params: RequestParams,
+    retryCount = 0
+  ): Promise<Stream.Readable> {
+    try {
+      return await this._request(params)
+    } catch (e) {
+      if (e instanceof Error && e.message === expiredSocketMessage) {
+        if (this.retry_expired_sockets && retryCount < 3) {
+          this.logger.debug({
```
mshustov: Maybe it should be `trace` level?

src/connection/adapter/base_http_adapter.ts
```diff
@@ -159,12 +193,88 @@ export abstract class BaseHttpAdapter implements Connection {
...
+              })
+              pipeStream()
+            }
+          } else {
+            const socketId = uuid.v4()
+            this.logger.trace({
+              module: 'Connection',
+              message: `Using a new socket ${socketId}`,
+            })
+            Object.assign(socket, { id: socketId })
```
mshustov: Could you answer the proposal here https://github.com/ClickHouse/clickhouse-js/pull/169#discussion_r1245571173 ?

src/connection/adapter/base_http_adapter.ts
```diff
@@ -159,12 +193,88 @@ export abstract class BaseHttpAdapter implements Connection {
...
+            this.known_sockets.set(socket, {
+              id: socketId,
+              last_used_time: Date.now(),
+            })
+            pipeStream()
+          }
+        } else {
+          // keep alive is disabled or retry mechanism is not enabled;
+          // no need to track the reused sockets
+          pipeStream()
```
mshustov, line 269: nit: to improve readability, we could move the short part higher

src/client.ts
slvrtrn: Yes

src/connection/adapter/base_http_adapter.ts
slvrtrn: Updated

src/connection/adapter/base_http_adapter.ts
slvrtrn: Thanks for pointing this out, this is rudimentary (we use WeakMap now).

__tests__/unit/http_adapter.test.ts
slvrtrn: Fixed, the issue was that I forgot to emit `socket` event that actually triggers the stream pipeline.

src/connection/adapter/base_http_adapter.ts
```diff
@@ -159,12 +193,87 @@ export abstract class BaseHttpAdapter implements Connection {
...
+              socket.destroy() // immediately terminate the connection
+              request.destroy()
+              reject(new Error(expiredSocketMessage))
+            } else {
+              this.logger.trace({
+                module: 'Connection',
+                message: `Socket ${socketInfo.id} is safe to be reused`,
+              })
+              this.known_sockets.set(socket, {
+                ...socketInfo,
```
mshustov: nit: we can make the intention explicit with 
```js
 this.known_sockets.set(socket, {
   id: socketInfo.id,
   last_used_time: Date.now(),
```

src/connection/adapter/base_http_adapter.ts
```diff
@@ -159,12 +193,87 @@ export abstract class BaseHttpAdapter implements Connection {
...
+          // no need to track the reused sockets;
+          // keep alive is disabled or retry mechanism is not enabled
+          pipeStream()
+        }
+
+        // this is for request timeout only.
+        // The socket won't be actually destroyed,
+        // and it will be returned to the pool.
+        // TODO: investigate if can actually remove the idle sockets properly
+        socket.setTimeout(this.config.request_timeout, onTimeout)
```
mshustov, line 276: Do we need `unref`, or is it already called under the hood?

mshustov (approved):

src/connection/adapter/base_http_adapter.ts
slvrtrn: Destroy removes the timeouts: https://github.com/nodejs/node/blob/main/lib/net.js#L796-L798

Issue #120, opened 2022-10-20
LZ4, BZ2, snappy compression support


slvrtrn:
Added in 22.10: https://github.com/ClickHouse/ClickHouse/pull/42071

slvrtrn:
LZ4 is blocked:
- https://github.com/pierrec/node-lz4 does not support modern Node.js versions
- https://github.com/antoniomuso/lz4-napi lacks streaming support

slvrtrn:
WIP branch (unmerged) https://github.com/ClickHouse/clickhouse-js/compare/main...lz4

Dhruv-Garg79:
This library can be used for zstd - https://github.com/mongodb-js/zstd
It outperforms the native one.

Issue #89, opened 2022-09-13
Add Nested data type support


slvrtrn:
See https://clickhouse.com/docs/en/sql-reference/data-types/nested-data-structures/nested

Fix the [failing test.](https://github.com/ClickHouse/clickhouse-js/blob/8a8ce2bde95ebc0b557bae1d1cba23c00b44405e/__tests__/integration/data_types.test.ts#L472-L523)

antoniovizuete:
Hi @slvrtrn!

I'm willing to work on adding support for nested data types.

My approach would be to flatten the nested structures, such as objects. So, I would choose the `Record<K, V>` TS type to support the `**Nested**`data type, because both have the same goal, to handle structured data.

If we assume the explanation, then we face incompatibility with the `Map(K, V)` data type. According to the **[[official docs](https://clickhouse.com/docs/en/integrations/language-clients/nodejs#supported-clickhouse-data-types)](https://clickhouse.com/docs/en/integrations/language-clients/nodejs#supported-clickhouse-data-types)**, the `Map(K, V)` data type is supported by the `Record<K, V>` TS type.

This means that I won’t be able to differentiate between `Map` and `Nested` when attempting to flatten them.

In my opinion, the `Map(K, V)` data type should be supported by the `Map<K, V>` TS type. There are a couple of advantages to supporting it as a TS Map:

- It would create semantic coherence between the Clickhouse data type and the TS type.
- The TS Map API is better suited for handling data map structures than the TS Record type.

What do you think about supporting Clickhouse Map with TS Map?

Many thanks

PR #174, opened 2023-07-10
Add support for URL connection string in ClickHouse client

xrip:
After clickhouse-client added connection string support (https://github.com/ClickHouse/ClickHouse/pull/50689) there is now nothing stops to add same support to this client.

## Summary
This pull request add support for URL connection string in ClickHouse client.

Extended the ClickHouse client configuration to accept a URL connection string which can contain user info, hosts and ports, database name and query parameters. This change provides flexible and standard connections setup. Updated related unit tests to reflect this change.

## Checklist

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -14,6 +14,10 @@ import type { ClickHouseSettings } from './settings'
 import type { InputJSON, InputJSONObjectEachRow } from './clickhouse_types'
 
 export interface ClickHouseClientConfigOptions {
+  /** ClickHouse connection string url
+   * ```clickhouse://[user_info@][hosts_and_ports][/dbname][?query_parameters]```
+   */
+  url?: string
   /** A ClickHouse instance URL.
    * <br/> Default value: `http://localhost:8123`. */
   host?: string
@@ -135,11 +139,13 @@ function validateConfig({ url }: NormalizedConfig): void {
   // TODO add SSL validation
 }
 
-function createUrl(host: string): URL {
+function createUrl(url: string): URL {
   try {
-    return new URL(host)
+    return new URL(url)
   } catch (err) {
-    throw new Error('Configuration parameter "host" contains malformed url.')
+    throw new Error(
+      'Configuration parameter "host" or "url" contains malformed url.'
+    )
   }
 }
 
@@ -158,9 +164,14 @@ function normalizeConfig(config: ClickHouseClientConfigOptions) {
       }
     }
   }
+
+  const url = config.url && createUrl(config.url)
+
   return {
     application_id: config.application,
-    url: createUrl(config.host ?? 'http://localhost:8123'),
+    url: createUrl(
+      config.host ?? (url && `http://${url.host}`) ?? 'http://localhost:8123'
+    ),
     request_timeout: config.request_timeout ?? 300_000,
     max_open_connections: config.max_open_connections ?? Infinity,
     tls,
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -135,11 +139,13 @@ function validateConfig({ url }: NormalizedConfig): void {
...
 
-function createUrl(host: string): URL {
+function createUrl(url: string): URL {
   try {
-    return new URL(host)
+    return new URL(url)
   } catch (err) {
-    throw new Error('Configuration parameter "host" contains malformed url.')
+    throw new Error(
+      'Configuration parameter "host" or "url" contains malformed url.'
```
mshustov: Let's refactor the function to report in what context it fails. for example:
```js
`Configuration parameter ${type} contains malformed url.`
````

File: src/client.ts [hunk 2]
```diff
@@ -14,6 +14,10 @@ import type { ClickHouseSettings } from './settings'
 import type { InputJSON, InputJSONObjectEachRow } from './clickhouse_types'
 
 export interface ClickHouseClientConfigOptions {
+  /** ClickHouse connection string url
+   * ```clickhouse://[user_info@][hosts_and_ports][/dbname][?query_parameters]```
+   */
+  url?: string
```
mshustov, line 20: Considering that `url` might conflict with other params (host, dbname, user, password, params, etc.) and it's easy to make a typo in a string-based configuratuion, I'm not sure it should be supported. This logic might be easily implemented in the application 
```js
const url = new URL('clickhouse://clickhouse://john:secret@127.0.0.1:9000')
const client = createClient({ host: url.host, .... });
```

src/client.ts
```diff
@@ -158,20 +164,28 @@ function normalizeConfig(config: ClickHouseClientConfigOptions) {
...
     tls,
     compression: {
       decompress_response: config.compression?.response ?? true,
       compress_request: config.compression?.request ?? false,
     },
-    username: config.username ?? 'default',
-    password: config.password ?? '',
-    database: config.database ?? 'default',
-    clickhouse_settings: config.clickhouse_settings ?? {},
+    username: config.username ?? (url && url.username) ?? 'default',
```
mshustov, line 181: I agree that configuration should take precedence, but I'd expect the client to throw an exception to ask an end user to specify the correct value explicitly.

mshustov (commented):
Considering that `url` might conflict with other params (host, dbname, user, password, params, etc.) and it's easy to make a typo in a string-based configuratuion, I'm not sure it should be supported. This logic might be easily implemented in the application 
```js
const url = new URL('clickhouse://clickhouse://john:secret@127.0.0.1:9000')
const client = createClient({ host: url.host, .... });
```

WDYT @slvrtrn

src/client.ts
xrip: I think it should be reflected in documentation, and not throw error cause it's not something that will break execution. It's also very flexible to have, for example following setup: 

```Javascript
const client = createClient({ url: 'clickhouse://john@127.0.0.1:9000', password: process.env.CH_PASSWORD  });
```
It's follows logic beside original CH behaviour (https://github.com/ClickHouse/ClickHouse/pull/50689)

src/client.ts
xrip: Ok, made error message more verbose and split unit test

slvrtrn:
Thanks for your contribution, @xrip, but I agree with @mshustov here.

This example:

```ts
const url = new URL('clickhouse://clickhouse://john:secret@127.0.0.1:9000')
const client = createClient({ host: url.host, .... });
```

is very simple and trivial to implement when needed in the application, so I don't see the necessity to build it into the client itself.

I also understand that having a PR declined does not feel good. To avoid situations like this in the future, I think we should start with a feature request issue, where an intention to contribute could be stated. This way, we can discuss the pros and cons in advance and decide if it's worth investing the time into writing the actual code/tests and reviewing it later.

I'd also like to add that we have a backlog of triaged issues if you are willing to contribute, for example:
- https://github.com/ClickHouse/clickhouse-js/issues/89 (last time we checked, there were some issues with inserting the data)
- https://github.com/ClickHouse/clickhouse-js/issues/120 (LZ4 streaming implementation is missing)
- https://github.com/ClickHouse/clickhouse-js/issues/150 (brainstorm/debugging is required)

Issue #191, opened 2023-07-20
Generate updated server settings typings


slvrtrn:
[Current version](https://github.com/ClickHouse/clickhouse-js/blob/730b1b2516e2d47dc9a32b1d8d0b8ba8ceb95ead/src/settings.ts) is outdated.
There were new settings introduced since Aug'22, for example, `async_insert_max_query_number`

Issue #181, opened 2023-07-13
Add browser client documentation and examples


slvrtrn:
Currently, we only have [Node.js docs](https://clickhouse.com/docs/en/integrations/language-clients/nodejs) and [examples](https://github.com/ClickHouse/clickhouse-js/tree/main/examples).

- [x] Browser documentation
- [x] Browser examples

slvrtrn:
Mostly covered by https://github.com/ClickHouse/clickhouse-docs/pull/1383 + generic examples in the project.
Once merged, I think we can close this.

PR #192, opened 2023-07-24
Prepare 0.2.0-beta2 release

### LINKED ISSUE #178: Don't throw in Ping command
It just throws if it fails to connect, which is a bit awkward.

We can either just return false on a failed attempt or change the method signature to be some variation of Either concept to allow further debugging on the application side:

```ts
type PingResult = {
  type: 'Success'
} | {
  type: 'Failure'
  error: Error
}
```

### LINKED ISSUE #179: Firefox tests for browser version
As of #165, only ChromeHeadless is used for Karma runs. 
We should add at least FirefoxHeadless to the list.

slvrtrn:
## Summary
- Rename `@clickhouse/client-browser` -> `@clickhouse/client-web` as it makes more sense, considering the supported platforms.
- Improve Firefox compatibility.
- Updated examples.
- Resolve #179 
- Resolve #178 - CORS won't work for the `/ping` endpoint, so for the web version, there is a workaround with `SELECT 1`

## Checklist


slvrtrn:
I think we have only this as leftovers for a "proper" 0.2.0:
- #181 
- #191

CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+## 0.2.0-beta2
+
+- `@clickhouse/client-browser` renamed to `@clickhouse/client-web` as it makes more sense,
+  considering the supported platforms.
+- Firefox is now officially supported with `@clickhouse/client-web`.
+
+### Breaking changes (both Node.js/Web)
```
mshustov, line 7: We should combine all the breaking changes in a single release note for `0.2.0` later.

packages/client-common/src/connection.ts
```diff
@@ -42,8 +42,14 @@ export interface ConnQueryResult<Stream> extends ConnBaseResult {
 export type ConnInsertResult = ConnBaseResult
 export type ConnExecResult<Stream> = ConnQueryResult<Stream>
 
+export type ConnPingResult =
```
mshustov, line 45: 🚀

mshustov (approved):

Issue #202, opened 2023-10-04
HTTP socket closes even when keep_alive is used


varrocs:
### Describe the bug

When using `insert()` with a `keep_alive` connection and the `async_insert` option the socket is closed after each insert request.

### Expected behaviour

The socket is kept open and it is reused for the next insert.

### Code example
```ts
const { createClient } = require('@clickhouse/client');

//create table test_table(`a` UInt16, `b` UInt16) ENGINE=MergeTree PRIMARY KEY a;

const TABLE = process.env.DOMAIN_REPORT_TABLE ?? 'test_table';

const client = createClient({
  host: process.env.CLICKHOUSE_HOST ?? 'http://localhost:18123',
  username: process.env.CLICKHOUSE_USER ?? 'default',
  password: process.env.CLICKHOUSE_PASSWORD,
  keep_alive: {
    enabled: true,
    socket_ttl: 600_000,
    retry_on_expired_socket: true,
  },
  async_insert: true, // I know this is redundant
  clickhouse_settings: {
    async_insert: true,
  },
  // To make sure inserts happen on the same connection
  max_open_connections: 1,
})


const insert = ( params ) => {
  void client.insert(
    {
      table: TABLE,
      format: 'JSONEachRow',
      values: params,
      async_insert: true, // I know, I know, redundancy
      clickhouse_settings: {
        async_insert: true,
      },
      keep_alive: {
        enabled: true,
        socket_ttl: 600_000,
        retry_on_expired_socket: true,
      },
    });
};

const wait = (ms) => new Promise((resolve) => setTimeout(resolve, ms) );

const main = async () => {
  await insert({a: 1, b: 1});
  await wait(1);
  await insert({a: 2, b: 2});
  await wait(1);
  await insert({a: 3, b: 3});
};

void main();
```

### Error log
See the tcpdump outputs below

#### Environment
* Client version: 0.2.2
* Language version: nodejs v20.6.0
* OS:  Ubuntu 22.04

#### ClickHouse server
* ClickHouse Server version: 23.9.1 revision 54466
* ClickHouse Server non-default settings, if any: -
* `CREATE TABLE` statements for tables involved: ```create table test_table(`a` UInt16, `b` UInt16) ENGINE=MergeTree PRIMARY KEY a;```

The script that ran the clickhouse server:

```
[...]

PR #203, opened 2023-10-11
Drain insert HTTP response

varrocs:
In case of an insert call drain the HTTP response instead of destroying it. 

## Summary
Calling destroy eventually closed the socket even if keep-alive was set. This was a performance limitation.
https://github.com/ClickHouse/clickhouse-js/issues/202

## Checklist

Note: I don't know how to test this

Note: I don't know if it applies to my change

### DIFF
--- a/packages/client-node/src/connection/node_base_connection.ts
+++ b/packages/client-node/src/connection/node_base_connection.ts
@@ -354,6 +354,40 @@ export abstract class NodeBaseConnection
     }
   }
 
+  async drainHttpResponse(stream: Stream.Readable): Promise<void> {
+    return new Promise((resolve, reject) => {
+      function dropdata() {
+        // We don't care about the data
+      }
+
+      function onEnd() {
+        removeListeners()
+        resolve()
+      }
+
+      function onError(err: Error) {
+        removeListeners()
+        reject(err)
+      }
+
+      function onClose() {
+        removeListeners()
+      }
+
+      function removeListeners() {
+        stream.removeListener('data', dropdata)
+        stream.removeListener('end', onEnd)
+        stream.removeListener('error', onError)
+        stream.removeListener('onClose', onClose)
+      }
+
+      stream.on('data', dropdata)
+      stream.on('end', onEnd)
+      stream.on('error', onError)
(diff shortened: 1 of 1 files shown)


packages/client-node/src/connection/node_base_connection.ts
```diff
@@ -354,6 +354,40 @@ export abstract class NodeBaseConnection
     }
   }
 
+  async drainHttpResponse(stream: Stream.Readable): Promise<void> {
+    return new Promise((resolve, reject) => {
+      function dropdata() {
+        // We don't care about the data
```
slvrtrn, line 360: Do we need to add `resolve()` here? Also, minor nit: `dropData`

slvrtrn:
> I don't know how to test this

If we want to test it, we need to spy on sockets somehow (with max connections set to 1); I am afraid there is no other way.

packages/client-node/src/connection/node_base_connection.ts
varrocs: > Do we need to add resolve() here?

No. This function will be called when data is available but we just want to throw away the data, we don't care.

varrocs: >minor nit: dropData

Fixed

varrocs:
> If we want to test it, we need to spy on sockets somehow

Well, let's say I don't insist on testing this.

varrocs:
> If we want to test it, we need to spy on sockets somehow

So it's me asking if you accept the pull request without writing tests to test if the consecutive requests are going on the same socket

slvrtrn:
@varrocs, I think we can live without such a test for now cause nothing else is broken, and we have a specific test for consecutive inserts with one socket, IIRC (and they pass).

I think I'd only ask you to confirm that with this change, the TCP dump looks like the fixed version mentioned here: https://github.com/ClickHouse/clickhouse-js/issues/202#issue-1926461538

varrocs:
Yes you have such test (with a comment that saying that how strange it is that the socket mostly reopened even with keep alive)
And yes, the tcpdump looks good.

Issue #171, opened 2023-07-05
Use query stream and get some broken symbols


AlekseyVerba:
### Describe the bug
I use query stream and get broken symbols.  In database data is normal, without any defects. When i get data from stream, i get something strange

i get these rows:
"[\"ᴺᵁᴸᴸ\", \"45049\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"Сайт\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺ���ᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"256\", \"0\", \"spb.evraz.market\", \"Каталог товаров\", \"Пользователь\", \"\\/metalloprokat\\/\", \"0.256\", \"Время ответа сайта\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\"]"

"[\"A89648DF371528B911EB7CC47A32031D\", \"45027\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"Проведе��ие заказа\", \"A8D548DF371528B911ED9BDB5E11ABCD\", \"ᴺᵁᴸᴸ\", \"ИН00-047642\", \"ИН00-040371 от 21.02.2022 16:43:08\", \"ᴺᵁᴸᴸ\", \"A8B948DF371528B911EC7D1294FB667E\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"1925\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"ᴺᵁᴸᴸ\", \"3.154\"]"

[...]

PR #204, opened 2023-10-16
[Node.js] Fix incomplete rows handling in streaming transformations

slvrtrn:
## Summary
* Resolves https://github.com/ClickHouse/clickhouse-js/issues/171
* Updates contribution guide
* Prepares 0.2.4

In specific scenarios, with a large number of incoming rows and/or large chunk sizes, incomplete rows were not handled correctly, and produced corrupted UTF-8 strings.

NB: I still need to check the web version, as I suspect it could have the same bug. See #205 

## Checklist

### DIFF
--- a/packages/client-node/__tests__/integration/node_streaming_e2e.test.ts
+++ b/packages/client-node/__tests__/integration/node_streaming_e2e.test.ts
@@ -114,7 +114,7 @@ describe('[Node.js] streaming e2e', () => {
     }
 
     describe('large amount of rows', () => {
-      fit('should work with .json()', async () => {
+      it('should work with .json()', async () => {
         const { table, values } = await generateData({
           rows: 10000,
           minWords: 5,
@@ -129,7 +129,7 @@ describe('[Node.js] streaming e2e', () => {
         expect(result).toEqual(values)
       })
 
-      fit('should work with .stream()', async () => {
+      it('should work with .stream()', async () => {
         const { table, values } = await generateData({
           rows: 10000,
           minWords: 5,
@@ -153,7 +153,7 @@ describe('[Node.js] streaming e2e', () => {
     })
 
     describe("rows that don't fit into a single chunk", () => {
-      fit('should work with .json()', async () => {
+      it('should work with .json()', async () => {
         const { table, values } = await generateData({
           rows: 5,
           minWords: 5000,
@@ -168,7 +168,7 @@ describe('[Node.js] streaming e2e', () => {
         expect(result).toEqual(values)
       })
 
-      fit('should work with .stream()', async () => {
+      it('should work with .stream()', async () => {
         const { table, values } = await generateData({
           rows: 5,
           minWords: 5000,
--- a/packages/client-node/__tests__/integration/node_streaming_e2e.test.ts
+++ b/packages/client-node/__tests__/integration/node_streaming_e2e.test.ts
@@ -1,6 +1,8 @@
 import type { Row } from '@clickhouse/client-common'
 import { type ClickHouseClient } from '@clickhouse/client-common'
+import { fakerRU } from '@faker-js/faker'
 import { createSimpleTable } from '@test/fixtures/simple_table'
+import { createTableWithFields } from '@test/fixtures/table_with_fields'
 import { createTestClient, guid } from '@test/utils'
 import Fs from 'fs'
 import split from 'split2'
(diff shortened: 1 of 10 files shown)


packages/client-node/src/result_set.ts
```diff
@@ -35,34 +38,64 @@ export class ResultSet implements BaseResultSet<Stream.Readable> {
...
+            chunkOffset = 0
+            currentDataSize = chunk.length
+          }
+          data.fill(chunk, chunkOffset, currentDataSize)
+          data.fill(0, currentDataSize, data.length)
+        } else {
+          // This is the first chunk, and we need to allocate enough memory
+          // to hold the chunk itself and any potential incomplete rows
+          currentDataSize = chunk.length
+          data = Buffer.alloc(chunk.length * 2, 0)
```
slvrtrn: Surprisingly, `allocUnsafe` without additional zero fills does not have any benefits here, at least not on the `hits_v1` dataset which is rather large (~9M rows).

packages/client-node/src/result_set.ts
```diff
@@ -35,34 +38,64 @@ export class ResultSet implements BaseResultSet<Stream.Readable> {
...
+                }
               },
             })
           } else {
             if (rows.length) {
               this.push(rows)
             }
-            break
+            // Buffer after currentDataSize index is filled with zeroes
+            incompleteRow = data.subarray(startIdx, currentDataSize)
```
slvrtrn: Incomplete rows need to stack, as we can even have a row that does not fit in a single chunk. I will also be adding these checks.

packages/client-node/src/result_set.ts
```diff
@@ -35,34 +38,64 @@ export class ResultSet implements BaseResultSet<Stream.Readable> {
 
     validateStreamFormat(this.format)
 
-    let decodedChunk = ''
+    let data: Buffer
+    let incompleteRow: Buffer | undefined
     const toRows = new Transform({
```
slvrtrn, line 42: Tests WIP. I need to generate a dataset, something with walls of text inside, that will be failing on the current main branch.

packages/client-node/src/result_set.ts
```diff
@@ -35,34 +38,64 @@ export class ResultSet implements BaseResultSet<Stream.Readable> {
...
+          if (incompleteRow !== undefined) {
+            chunkOffset = incompleteRow.length
+            currentDataSize = chunkOffset + chunk.length
+            data.fill(incompleteRow, 0, chunkOffset)
+          } else {
+            chunkOffset = 0
+            currentDataSize = chunk.length
+          }
+          data.fill(chunk, chunkOffset, currentDataSize)
+          data.fill(0, currentDataSize, data.length)
```
slvrtrn: Maybe there is a more efficient way to work with these buffers. What is more interesting is that this implementation is 5-7% slower than the main branch (bugged) one. I was expecting it to be the other way around.

packages/client-node/src/result_set.ts
```diff
@@ -35,34 +38,64 @@ export class ResultSet implements BaseResultSet<Stream.Readable> {
...
       transform(
         chunk: Buffer,
-        encoding: BufferEncoding,
+        _encoding: BufferEncoding,
         callback: TransformCallback
       ) {
-        decodedChunk += chunk.toString()
+        let currentDataSize
+        // The buffer was already allocated
+        // x2 chunk size should be enough for any leftover incomplete rows
```
slvrtrn: No, it is actually not enough. I will add ad-hoc resizing cause we can have huge strings or other data with variable length out of nowhere in some chunks.

packages/client-node/__tests__/integration/node_streaming_e2e.test.ts
```diff
@@ -73,4 +75,114 @@ describe('[Node.js] streaming e2e', () => {
...
+      table: string
+      values: { id: number; sentence: string; timestamp: string }[]
+    }> {
+      const table = await createTableWithFields(
+        client as ClickHouseClient,
+        `sentence String, timestamp String`
+      )
+      const values = [...new Array(rows)].map((_, id) => ({
+        id,
+        // it seems that it is easier to trigger an incorrect behavior with non-ASCII symbols
```
mshustov, line 99: 👍 +1

packages/client-node/src/result_set.ts
```diff
@@ -35,33 +38,57 @@ export class ResultSet implements BaseResultSet<Stream.Readable> {
...
-              text,
-              json<T>(): T {
-                return decode(text, 'JSON')
-              },
-            })
+        let lastIdx = 0
+        // first pass on the current chunk
+        // using the incomplete row from the previous chunks
+        let idx = chunk.indexOf(NEWLINE)
+        if (idx !== -1) {
```
mshustov, line 53: Do you think we could extract some of the logic into a separate piece? It might be a little bit hard to reason about

mshustov (approved):

packages/client-node/src/result_set.ts
slvrtrn: I will check what's possible in one of the follow-ups.

Issue #164, opened 2023-06-04
Not working if the http interface is reverse-proxied by nginx with a suffix


ngugcx:
### Describe the bug
The http interface of my clickhouse server is reverse-proxied using nignx with suffix.
http://localhost:8123 is mapped to https://my.domain/clickhouse

```
This error is prompted when client.ping():
node:internal/process/promises:288
            triggerUncaughtException(err, true /* fromPromise */);
            ^

Error: Client network socket disconnected before secure TLS connection was established
    at connResetException (node:internal/errors:717:14)
    at TLSSocket.onConnectEnd (node:_tls_wrap:1595:19)
    at TLSSocket.emit (node:events:525:35)
    at endReadableNT (node:internal/streams/readable:1359:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  code: 'ECONNRESET',
  path: null,
  host: 'my.domain',
  port: 443,
  localAddress: undefined
}
```

It's working when http://localhost:8123 is mapped to https://my.domain/ without a suffix.
And when accessing https://my.domain/clickhouse using curl, all works well too.

### Steps to reproduce
1. Setup reverse proxy using nignx, map http://localhost:8123 to https://my.domain/clickhouse
2. execute ping()

### Expected behaviour

ping() should return true.

### Code example

```
const client = createClient({
  host: process.env.CLICKHOUSE_HOST ?? 'https://my.domain/clickhouse',
  username: process.env.CLICKHOUSE_USER ?? 'username',
  password: process.env.CLICKHOUSE_PASSWORD ?? 'password'
})

client.ping()
```

#### Environment
* Client version: 0.0.16
* Language version: node v18.16.0
* OS: Windows 10

#### ClickHouse server
* ClickHouse Server version: 23

slvrtrn:
I think it is because the HTTP connection is chosen instead of HTTPS: https://github.com/ClickHouse/clickhouse-js/blob/main/src/connection/connection.ts#L74-L81

slvrtrn:
@martoche, you are referring to an issue with the web version; is it correct? 
I think we could add an optional `pathname` (or similar) to the client config itself.

For example:

```ts
const client = createClient({
  pathname: '/clickhouse'
})

// ### Node.js
// ping -> /clickhouse/ping
// other -> /clickhouse

// ### Web (no specific ping call, it just uses SELECT 1)
// all requests -> /clickhouse
```

martoche:
[...]

PR #207, opened 2023-10-30
Don't override `pathname` in `transformUrl`

slvrtrn:
## Summary
See the [discussion](https://github.com/ClickHouse/clickhouse-js/issues/164#issuecomment-1785166626). Before these changes, `host` such as `http://proxy:8123/clickhouse` will always lose its `/clickhouse` pathname segment after the `transformUrl` call. Now, it is preserved.

## Checklist

### DIFF
--- a/packages/client-common/__tests__/unit/transform_url.test.ts
+++ b/packages/client-common/__tests__/unit/transform_url.test.ts
@@ -45,6 +45,14 @@ describe('transformUrl', () => {
     expect(newUrl.toString()).toBe('http://clickhouse.com/clickhouse/foobar')
   })
 
+  it('allows a trailing slash in the pathname', () => {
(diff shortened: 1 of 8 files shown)


packages/client-common/__tests__/unit/transform_url.test.ts
```diff
@@ -20,6 +36,15 @@ describe('transformUrl', () => {
     expect(newUrl.toString()).toBe('http://clickhouse.com/foo')
   })
 
+  it('attaches pathname to an existing pathname', () => {
+    const url = new URL('http://clickhouse.com/clickhouse')
```
mshustov, line 40: could you add a test that it handles a trailing slash in the URL?

 ```js
const url = new URL('http://clickhouse.com/clickhouse/')
````

mshustov (approved):

PR #214, opened 2023-12-06
Expose `X-ClickHouse-Summary` header info

### LINKED ISSUE #210: There is no way to get number of inserted rows
### Describe the bug
When I insert rows in some client
```sql
INSERT INTO device_values(*)
   WITH
    (
        ....
    ) AS device_id
    SELECT ....
    WHERE device_id != 0
```
I can see result like
```
1 row(s) modified.
```
As the result value can be inserted or vice versa - not inserted. 
I did not found any way how to handle it in `clickhouse-js`. "1 row" is what I want to get after execution. I could check before insert and did not make insert but it would be not so atomic operation.

slvrtrn:
## Summary
* Adds `summary` field to `insert`/`exec`/`command` result, which is parsed from the `X-ClickHouse-Summary` response header.
* Geo data type test is temporarily marked as ignored as it is still unclear why it fails with a cluster setup on the latest master.

Node.js only. The web version seems to have some CORS limitations.

Question: shall we expose `summary` for `query`, too? It is very unlikely to be correct with streaming scenarios, and for non-streaming ones, we know the number of rows anyway.

Resolves #210 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,11 @@
-## 0.2.7 (Web only)
+## 0.2.7 (Common, Node.js, Web)
+
+### New features
+
+- (Node.js only) `X-ClickHouse-Summary` response header is now parsed when working with `insert`/`exec`/`command` methods.
+  See the [related test](./packages/client-node/__tests__/integration/node_summary.test.ts) for more details.
+  NB: it is guaranteed to be correct only for non-streaming scenarios.
+  Web version does not currently support this due to CORS limitations. ([#210](https://github.com/ClickHouse/clickhouse-js/issues/210))
 
 ### Bug fixes
 
--- a/packages/client-common/__tests__/integration/data_types.test.ts
+++ b/packages/client-common/__tests__/integration/data_types.test.ts
@@ -401,7 +401,9 @@ describe('data types', () => {
     ).toEqual('3\n')
   })
 
-  it('should work with geo', async () => {
+  // FIXME: somehow broken after, probably, https://github.com/ClickHouse/ClickHouse/pull/56724
+  //  enable once resolved.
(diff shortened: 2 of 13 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
-## 0.2.7 (Web only)
+## 0.2.7 (Common, Node.js, Web)
+
+### New features
+
+- (Node.js only) `X-ClickHouse-Summary` response header is now parsed when working with `insert`/`exec`/`command` methods.
```
mshustov, line 5: don't we need `send_progress_in_http_headers=1` header? see https://github.com/ClickHouse/clickhouse-connect/blob/6928aed5845c944007b6bf93e806313a96a93fbc/clickhouse_connect/driver/httpclient.py#L368-L370

slvrtrn: Interesting, something to discuss for sure. In tests, for non-streaming operations, it works regardless.

mshustov: According to the [Changelog](https://github.com/ClickHouse/ClickHouse/blob/b05a3d02ed5bbd8304bc48763a7c58b6f419eb6d/docs/changelogs/v22.6.1.1985-stable.md#improvement), it should work for non-streaming operations

packages/client-common/__tests__/integration/data_types.test.ts
```diff
@@ -401,7 +401,9 @@ describe('data types', () => {
     ).toEqual('3\n')
   })
 
-  it('should work with geo', async () => {
+  // FIXME: somehow broken after, probably, https://github.com/ClickHouse/ClickHouse/pull/56724
```
mshustov: The PR is merged

Issue #217, opened 2023-12-23
Allow to specify a list of columns for `insert`


cjroebuck:
### Describe the bug

Not sure if this is a clickhouse-js specific bug or just clickhouse in general. 

Whenever I try to insert JSON as a raw `String` with `EPHEMERAL` column and then have derived columns using various JSONExtract* functions, I can't seem to get it to work. I'm using the example from the JSON docs [here](https://clickhouse.com/docs/en/integrations/data-formats/json#adding-primary-keys).

### Steps to reproduce

Run the following code to setup tables and insert test rows.

```
import { createClient } from "@clickhouse/client"; 

void (async () => {
  const client = createClient({
    clickhouse_settings: {
      date_time_input_format: "best_effort",
      allow_experimental_object_type: 1,
    },
  });

  await client.command({
    query: `DROP TABLE IF EXISTS github_json_ephemeral`,
  });

  await client.command({
    query: `CREATE table github_json_ephemeral
  (
     event_type LowCardinality(String) DEFAULT JSONExtractString(message_raw, 'type'),
     repo_name LowCardinality(String) DEFAULT JSONExtractString(message_raw, 'repo.name'),
     message JSON DEFAULT message_raw,
     message_raw String EPHEMERAL
  ) ENGINE = MergeTree ORDER BY (event_type, repo_name)`,
  });

  await client.insert({
    table: "github_json_ephemeral",
    format: "JSON",
    values: [
      {
        message_raw: {
          type: "PushEvent",
          created_at: "2022-01-04 07:00:00",
          actor: {
            avatar_url: "https://avatars.githubusercontent.com/u/41898282?",
            display_login: "github-actions",
            gravatar_id: "",
            id: 41898282,
            login: "github-actions[bot]",
            url: "https://api.github.com/users/github-actions[bot]",
          },
          repo: {
            id: 410071248,
            name: "pioug/yield-data",
            url: "https://api.github.com/repos/pioug/yield-data",
          },
        },
      },
    ],
  });
})
```

### Expected behaviour

Output:

```
rows [
  {
    event_type: 'PushEvent',
    repo_name: 'pioug/yield-data',
    message: {
      actor: [Object],
      created_at: '2022-01-04 07:00:00',
      repo: [Object],
      type: 'PushEvent'
    },
  }
```

### Actual behaviour

The derived columns are blank and the JSON column is empty:
```
rows [
  { event_type: '', repo_name: '', message: { _dummy: 0 } }
]
```

[...]

PR #218, opened 2023-12-23
Allow to specify or exclude insert columns

slvrtrn:
## Summary

It is now possible to either specify a list of columns to insert the data into or a list of excluded columns:

```ts
// Generated query: INSERT INTO mytable (message) FORMAT JSONEachRow
await client.insert({
  table: 'mytable',
  format: 'JSONEachRow',
  values: [{ message: 'foo' }],
  columns: ['message'],
})

// Generated query: INSERT INTO mytable (* EXCEPT (message)) FORMAT JSONEachRow
await client.insert({
  table: 'mytable',
  format: 'JSONEachRow',
  values: [{ id: 42 }],
  columns: { exclude: ['message'] },
})
```

Added more examples:
- [Including specific columns or excluding certain ones instead](https://github.com/ClickHouse/clickhouse-js/blob/b55c22243a4a30767a3ab9770cedb947fd37f093/examples/insert_exclude_columns.ts)
- [Leveraging this feature](https://github.com/ClickHouse/clickhouse-js/blob/b55c22243a4a30767a3ab9770cedb947fd37f093/examples/insert_ephemeral_columns.ts) when working with [ephemeral columns](https://clickhouse.com/docs/en/sql-reference/statements/create/table#ephemeral)

Resolves https://github.com/ClickHouse/clickhouse-js/issues/217

NB: the `EXCEPT` clause works with ephemeral types rather strangely (see the tests). It is also possible to exclude every single column via HTTP, and the query does not fail, though the data is not inserted - I couldn't simulate that via CLI as it does not allow empty values. It is also possible to exclude a column that does not exist (same in the CLI). All of this is to be investigated.

## Checklist
- [ ] Website docs update

### DIFF
--- a/packages/client-common/__tests__/integration/insert_specific_columns.test.ts
+++ b/packages/client-common/__tests__/integration/insert_specific_columns.test.ts
@@ -1,0 +1,387 @@
+import { type ClickHouseClient } from '@clickhouse/client-common'
+import { createTableWithFields } from '@test/fixtures/table_with_fields'
+import { createTestClient } from '../utils'
+
+describe('Insert with specific columns', () => {
+  let client: ClickHouseClient
+  let table: string
+
+  beforeEach(async () => {
+    client = createTestClient({
+      clickhouse_settings: {
+        allow_experimental_object_type: 1,
+      },
+    })
+  })
+  afterEach(async () => {
+    await client.close()
+  })
+
+  // Inspired by https://github.com/ClickHouse/clickhouse-js/issues/217
+  // Specifying a particular insert column is especially useful with ephemeral types
+  describe('list of columns', () => {
+    beforeEach(async () => {
+      // `message_raw` will be used as a source for default values here
+      table = await createTableWithFields(
+        client,
+        `
+        event_type  LowCardinality(String) DEFAULT JSONExtractString(message_raw, 'type'),
+        repo_name   LowCardinality(String) DEFAULT JSONExtractString(message_raw, 'repo', 'name'),
+        message     JSON                   DEFAULT message_raw,
+        message_raw String                 EPHEMERAL
+      ` // `id UInt32` will be added as well
+      )
+    })
+
+    it('should work with a single column', async () => {
+      await client.insert({
+        table,
+        values: [
+          {
+            message_raw: {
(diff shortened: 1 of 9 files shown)


packages/client-common/__tests__/integration/insert_specific_columns.test.ts
```diff
@@ -0,0 +1,387 @@
+import { type ClickHouseClient } from '@clickhouse/client-common'
+import { createTableWithFields } from '@test/fixtures/table_with_fields'
+import { createTestClient } from '../utils'
+
+describe('Insert with specific columns', () => {
```
mshustov, line 5: could you add examples with nested data as it' done in https://github.com/ClickHouse/clickhouse-connect/blob/526cc8b12a23b36ac1f846180d39fe1c04b263d7/examples/insert_examples.py

slvrtrn: I need to finally have a look at what's up with Nested via HTTP: https://github.com/ClickHouse/clickhouse-js/issues/89

packages/client-common/src/client.ts
```diff
@@ -229,10 +249,21 @@ export class ClickHouseClient<Stream = unknown> {
    */
   async insert<T>(params: InsertParams<Stream, T>): Promise<InsertResult> {
     const format = params.format || 'JSONCompactEachRow'
-
     this.valuesEncoder.validateInsertValues(params.values, format)
-    const query = `INSERT INTO ${params.table.trim()} FORMAT ${format}`
 
+    let columnsPart = ''
+    if (params.columns !== undefined) {
```
mshustov: optional nit: We could consider extractimg in a separate function to simplify testing:
```js
function getColumnsForInsert(columns) {
  if (!columns) return '';
  if (Array.isArray(columns) && columns.length > 0) {
        return `(${columns.join(',')})`
  if (
        isInsertColumnsExcept(columns) &&
        columns.except.length > 0
      ) {
        return` (* EXCEPT (${columns.except.join(',')}))`
      }
    }
  throw new Error(`unexpected value of columns: ${columns}`)
}
```

mshustov (approved):

Issue #211, opened 2023-11-22
Clickhouse client working with bun js


DeniroBlack:
Bun js is fasted speed js runtime runner

https://bun.sh/

But client not working. Difference with node in this file

client/dist/connection/node_http_connection.js:19

May be difference in http client

Please fix this in next versions. Or tell please how I may fix this myself

slvrtrn:
I am not familiar with this tool. However, I see `fetch` usage there. Does it work with `@clickhouse/client-web`?

DeniroBlack:
> I am not familiar with this tool. However, I see `fetch` usage there. Does it work with `@clickhouse/client-web`?

With web version works fine! Thanks!

dsumer:
Would love to see this issue being reopened. It's working with the `client-web` library, but that means we lose the ability to work with the connection pool and `keepAlive` feature which is only available in the `client` library.

Would be awesome if the `client` library would also be compatible with Bun, thank you! :)

slvrtrn:
@dsumer, I suspect the Node.js version will not work with Bun unless we rewrite it entirely to the fetch API. I think the right approach is to make keep-alive configurable in the web version instead - it will be there after https://github.com/ClickHouse/clickhouse-js/pull/220.

slvrtrn:
@dsumer, as 0.2.8 is out, it is now possible to modify Keep-Alive setting for the web version (and it is enabled by default). See the [release notes](https://github.com/ClickHouse/clickhouse-js/releases/tag/0.2.8).

PR #220, opened 2023-12-28
Make keep_alive configurable in the Web version

### LINKED ISSUE #186: Browser client KeepAlive setting
As of #165, KeepAlive is hardcoded to be disabled for browser implementation.
We need to check whether we will have the same issues as #150 if we enable it.

slvrtrn:
## Summary

Resolves #186. Related to https://github.com/ClickHouse/clickhouse-js/issues/211#issuecomment-1871032892.
KeepAlive was disabled for the web version without the possibility to re-enable it. 

## Checklist

The changelog and the docs will be updated after #218

### DIFF
--- a/packages/client-web/__tests__/integration/web_connection.test.ts
+++ b/packages/client-web/__tests__/integration/web_connection.test.ts
@@ -1,0 +1,42 @@
+import { createClient } from '../../src'
+import type { WebClickHouseClient } from '../../src/client'
+
+describe('[Web] Connection', () => {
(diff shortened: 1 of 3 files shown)


mshustov (approved):
I'm not sure it's necessary since keepalive in fetch has a different meaning https://developer.mozilla.org/en-US/docs/Web/API/fetch#keepalive 
OTOH, I can imagine that `keepalive` in Web API compatible env works as expected, so let's merge

PR #224, opened 2024-01-17
feat: added additional headers to clickhouse requests

teawithfruit:
## Summary
This PR enables clickhouse-js to use additional headers in http(s) requests. On my side it was necessary to be able to use a reverse proxy with basic auth in front. For example, something like this is possible from now on.

```ts
createClient({
  host: 'https://log.example.com',
  username: 'basic_auth_user',
  password: 'basic_auth_password',
  additionalHeaders: {
    'X-ClickHouse-User': 'clickhouse_user',
    'X-ClickHouse-Key': 'clickhouse_password',
  },
});
```

## Checklist

### DIFF
--- a/packages/client-common/src/client.ts
+++ b/packages/client-common/src/client.ts
@@ -92,7 +92,7 @@ export interface ClickHouseClientConfigOptions<Stream> {
     level?: ClickHouseLogLevel
   }
   session_id?: string
-  additionalHeaders?: object
+  additional_headers?: Record<string, number | string | string[]>
 }
 
 export type BaseClickHouseClientConfigOptions<Stream> = Omit<
@@ -336,7 +336,7 @@ function getConnectionParams<Stream>(
         : new DefaultLogger(),
       config.log?.level
     ),
-    additionalHeaders: config.additionalHeaders ?? {},
+    additional_headers: config.additional_headers,
   }
 }
 
--- a/packages/client-common/src/client.ts
+++ b/packages/client-common/src/client.ts
@@ -92,6 +92,7 @@ export interface ClickHouseClientConfigOptions<Stream> {
     level?: ClickHouseLogLevel
   }
   session_id?: string
+  additionalHeaders?: object
 }
 
 export type BaseClickHouseClientConfigOptions<Stream> = Omit<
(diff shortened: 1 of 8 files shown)


packages/client-common/src/client.ts
```diff
@@ -92,6 +92,7 @@ export interface ClickHouseClientConfigOptions<Stream> {
     level?: ClickHouseLogLevel
   }
   session_id?: string
+  additionalHeaders?: object
```
slvrtrn: Let's keep it in snake_case as the rest, and change the type to `Record<string, number | string | string[]>` to be more in line with the OutgoingHttpHeaders type.

packages/client-common/src/client.ts
```diff
@@ -335,6 +336,7 @@ function getConnectionParams<Stream>(
         : new DefaultLogger(),
       config.log?.level
     ),
+    additionalHeaders: config.additionalHeaders ?? {},
```
slvrtrn: It can be just `additional_headers?` without `??` as it is in `connection.ts`, i.e. we are ok with an undefined.

slvrtrn (changes requested):
Can you please 
- add these options to the web version as well
- add tests - for the web version, the suitable place likely will be in `packages/client-web/__tests__/integration/web_connection.test.ts` (fetch stubs are already there, should be easy to assert on headers); for Node.js, a similar one can be created as `packages/client-node/__tests__/integration/node_connection.test.ts` as we want to test the entire interaction (client options -> connection -> request). There are also examples of [http.request stubs here](https://github.com/ClickHouse/clickhouse-js/blob/20bd2868a2b5caf372fc501431c1bafb7d9a9758/packages/client-node/__tests__/unit/node_http_adapter.test.ts#L23-L24).

slvrtrn:
@teawithfruit, thanks for the contribution. I took the liberty to do some fixups in the tests, plus I noticed that `Record<string, number | string | string[]>` was a bit too permissive (`fetch` API is a bit more strict on that), so I reduced it to `Record<string, string>` as it matches both APIs.

Released as [0.2.9](https://github.com/ClickHouse/clickhouse-js/releases/tag/0.2.9).

teawithfruit:
Thanks for reacting such fast. 🙂

PR #225, opened 2024-01-26
Add default HTTP settings + readonly user switch

slvrtrn:
## Summary

See CHANGELOG. It is a breaking change for read-only users, but IMO necessary, as timeout issues when an LB terminates a long-running query connection are incredibly obscure, and this should at least cover AWS LB defaults.

It works similarly to how [clickhouse-connect](https://github.com/ClickHouse/clickhouse-connect/blob/61d6a1b2855b32fa9e0f050e6113cb29825cac9b/clickhouse_connect/driver/httpclient.py#L151-L159) handles it (but currently does not normalize it based on the request timeout).

## Checklist
- [x] Docs will be updated if we decide to merge this.


mshustov:
>It is a breaking change for read-only users,

Not for all readonly users, only for `readonly=1`? https://clickhouse.com/docs/en/operations/settings/constraints-on-settings

CHANGELOG.md
```diff
@@ -1,3 +1,33 @@
+## 0.2.10 (Common, Node.js, Web)
+
+### Breaking changes
```
mshustov, line 3: since it contains a potential breaking change, shouldn't it be `v0.3.0`?

CHANGELOG.md
```diff
@@ -1,3 +1,33 @@
...
+    response: false,
+  },
+})
+```
+
+With `send_progress_in_http_headers` and `http_headers_progress_interval_ms` settings now enabled by default, this is no longer sufficient. If you need to create a client instance for a read-only user, consider this instead:
+
+```ts
+const client = createClient({
+  readonly: true,
```
mshustov, line 25: in clickhouse, readonly has a bit different semantics: 0, 1, 2. let's align

CHANGELOG.md
```diff
@@ -1,3 +1,33 @@
+## 0.2.10 (Common, Node.js, Web)
+
+### Breaking changes
+
```
mshustov, line 4: Is it a breaking change not for all `readonly` users, only for `readonly=1`? https://clickhouse.com/docs/en/operations/settings/constraints-on-settings

slvrtrn:
@mshustov, I updated the version, docs, and CHANGELOG, plus added an example, WDYT?

mshustov (approved):

PR #227, opened 2024-02-11
Add tests for Nested data type

auvred:
## Summary

Closes https://github.com/ClickHouse/clickhouse-js/issues/89

It looks like the `Nested` data type is already supported, but the documentation says it isn't yet supported: https://clickhouse.com/docs/en/integrations/language-clients/javascript#supported-clickhouse-data-types

https://github.com/ClickHouse/clickhouse-docs/blob/d99f31aa0e8aa50085a737ca3402f09931c75b9f/docs/en/integrations/language-clients/js.md?plain=1#L739

This may lead to confusion.

I'm adding tests to prove that it does work, so I can update the status of support in the docs

## Checklist

### DIFF
--- a/packages/client-common/__tests__/integration/data_types.test.ts
+++ b/packages/client-common/__tests__/integration/data_types.test.ts
@@ -519,26 +519,21 @@ describe('data types', () => {
     await insertAndAssert(table, values)
   })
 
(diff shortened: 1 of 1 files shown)


slvrtrn:
Thanks for the contribution!

Issue #231, opened 2024-02-26
Design Bug: CLICKHOUSE_HOST is an incorrect variable name


iMerica:
[See RFC-1738](https://datatracker.ietf.org/doc/html/rfc1738)

Hi there, 

Thank you for building this library. This is just a quick bug report that  `CLICKHOUSE_HOST` is a poorly named variable name.  The host and hostname of a URL are very clear and unambiguous concepts. They typically refer only to the host where a server can be resolved. Not to the entire URI. If you want the entire URI, then call the variable `CLICKHOUSE_URI` or better yet call it `CLICKHOUSE_URL` and accept a connection string. 

See the host in the URL api of Javascript

```js
new URL("https://github.com").host
'github.com'
```
Also look how PostgreSQL uses the noun "host" to only refer to the hostname

https://www.postgresql.org/docs/6.4/jdbc19100.htm

slvrtrn:
`CLICKHOUSE_HOST` is not used in the code; it's from the examples, right?
Minimal URL support will be added soon in the scope of #228, and the `host` client configuration parameter will be deprecated.

PR #232, opened 2024-02-29
Add support for URL parameters parsing

### LINKED ISSUE #228: Add connection url support to the client
### Use case

We set database environment variables based on the standard format of xxx://[userspec@][hostspec][/dbname][?paramspec]
Clickhouse js client currently doesn't support it and we need to specify the env variables inconsistently to other databases.

### Describe the solution you'd like

Add support for connection string like clickhouse://[user_info@][hosts_and_ports][/dbname][?query_parameters]
as it is already supported in clickhouse https://github.com/ClickHouse/ClickHouse/pull/50689

### Additional context

ClickHouseClientConfigOptions defines the connection options

slvrtrn:
## Summary

Into the 1.0.0 branch.

Resolves #228 #231 

## Checklist

The docs will be updated once 1.0.0 is out.


CHANGELOG.md
```diff
@@ -32,6 +38,57 @@ NB: this is not necessary if a user has `READONLY = 2` mode as it allows to modi
 
 See also: [readonly documentation](https://clickhouse.com/docs/en/operations/settings/permissions-for-queries#readonly).
 
+### New features
+
+- Added `url` configuration parameter. It is intended to replace the deprecated `host`, which was already supposed to be passed a valid URL.
+- It is now possible to configure most of the client instance parameters with a URL. The URL format is `http[s]://[username:password@]hostname:port[/database][?param1=value1&param2=value2]`. The name of a particular parameter is supposed to reflect its path in the config options interface. The following parameters are supported:
+
+  - `readonly` - boolean. See below [1].
```
mshustov: optional: what if we use a table representation? 

| name | type |
| - | - | 
| readonly[1] | boolean |
| application_id | string |

CHANGELOG.md
```diff
@@ -32,6 +38,57 @@ NB: this is not necessary if a user has `READONLY = 2` mode as it allows to modi
...
+  - `application_id` - non-empty string.
+  - `session_id` - non-empty string.
+  - `request_timeout` - non-negative number.
+  - `max_open_connections` - non-negative number, greater than zero.
+  - `compression_request` - boolean.
+  - `compression_response` - boolean.
+  - `log_level` - allowed values: `OFF`, `TRACE`, `DEBUG`, `INFO`, `WARN`, `ERROR`.
+  - `keep_alive_enabled` - boolean.
+  - `clickhouse_settings_*` - see below [2].
```
mshustov: Nit: maybe `clickhouse_setting_*` since it's a single setting? a side note: python client uses `ch_` convention https://github.com/ClickHouse/clickhouse-connect/blob/01dcaf7c3494614e6c773ee10f744ff475d5b26e/clickhouse_connect/driver/__init__.py#L104

CHANGELOG.md
```diff
@@ -32,6 +38,57 @@ NB: this is not necessary if a user has `READONLY = 2` mode as it allows to modi
...
+  - `additional_headers_*` - see below [3].
```
mshustov: Should it be `http_header`? Are we going to support these settings for Native protocol?

CHANGELOG.md
```diff
@@ -32,6 +38,57 @@ NB: this is not necessary if a user has `READONLY = 2` mode as it allows to modi
...
+
+```ts
+createClient({
+  additional_headers: {
+    'x-clickhouse-auth': 'foobar',
+  },
+})
+```
+
+URL will _always_ overwrite the hardcoded values and a warning will be logged in this case.
```
mshustov:
```suggestion
*URL will _always_ overwrite the hardcoded values and a warning will be logged in this case.*
```

CHANGELOG.md
```diff
@@ -32,6 +38,57 @@ NB: this is not necessary if a user has `READONLY = 2` mode as it allows to modi
 
 See also: [readonly documentation](https://clickhouse.com/docs/en/operations/settings/permissions-for-queries#readonly).
 
+### New features
+
+- Added `url` configuration parameter. It is intended to replace the deprecated `host`, which was already supposed to be passed a valid URL.
```
mshustov: That is an excellent summary. Don't forget to add it to the client documentation.

packages/client-common/src/config.ts
```diff
@@ -0,0 +1,410 @@
...
+
+  let clickHouseSettings: ClickHouseSettings
+  let compressionSettings: CompressionSettings
+  // TODO: maybe validate certain settings that cannot be modified with read-only user
+  if (!config.readonly) {
+    clickHouseSettings = {
+      ...DefaultClickHouseSettings,
+      ...config.clickhouse_settings,
+    }
+    compressionSettings = {
```
mshustov, line 204: We need some tests here, the logic is not complex but hard to understand at first glance

mshustov (approved):

packages/client-common/src/config.ts
slvrtrn: sure, tests are WIP, will merge only after everything is tested.

CHANGELOG.md
slvrtrn: My idea was that it is supposed to reflect the path in the settings. We can add a shorthand, too, i.e., you can write `clickhouse_settings_` or just `ch_`

CHANGELOG.md
slvrtrn: Let's also rename and deprecate the existing `additional_headers`.

PR #238, opened 2024-03-11
Infer ResultSet type hints based on DataFormat

slvrtrn:
## Summary
Into 1.0.0. See the changelog entry. Docs will be updated accordingly.

[Screencast from 2024-03-09 08-10-26.webm](https://github.com/ClickHouse/clickhouse-js/assets/3175289/b66afcb2-3a10-4411-af59-51d2754c417e)

## Checklist


packages/client-node/src/index.ts
```diff
@@ -1,6 +1,7 @@
+export type { NodeClickHouseClient } from './client'
```
slvrtrn: Maybe it's better to re-export as just `ClickHouseClient`; WDYT @mshustov?

packages/client-node/src/result_set.ts
```diff
@@ -1,34 +1,67 @@
...
+import type { Readable, TransformCallback } from 'stream'
 import Stream, { Transform } from 'stream'
 import { getAsText } from './utils'
 
 const NEWLINE = 0x0a as const
 
-export class ResultSet implements BaseResultSet<Stream.Readable> {
+/** {@link Stream.Readable} with additional types for the `on(data)` method.
+ * Everything else is an exact copy from stream.d.ts */
+export type StreamReadable<T> = Omit<Stream.Readable, 'on'> & {
```
mshustov, line 18: Why do we need the custom definition?

packages/client-node/src/index.ts
mshustov: You export it at the bottom, don't you? if you ask about a type - yes, we can re-export it with a more generic name to shorten the name

packages/client-common/src/result.ts
```diff
@@ -1,4 +1,40 @@
...
+import type { ResponseJSON } from './clickhouse_types'
+import type {
+  DataFormat,
+  RawDataFormat,
+  RecordsJSONFormat,
+  SingleDocumentJSONFormat,
+  StreamableJSONDataFormat,
+} from './data_formatter'
+
+export type ResultJSONType<T, F extends DataFormat | unknown> =
```
mshustov, line 24: Maybe it is too much, but we might need some tests for such complex types. Or do you think the current JS tests in TypeScript cover all the cases?  https://effectivetypescript.com/2022/05/28/eslint-plugin-expect-type/

mshustov (approved):

packages/client-node/src/result_set.ts
slvrtrn: > additional types for the `on(data)` method.

By default, it's any there.

packages/client-common/src/result.ts
slvrtrn: I think this is a good idea, it's not too much to test - basically, a few calls for every DataFormat, and that's it. Could be put into some unit test that expects nothing, and just serves as an ESLint check.

packages/client-node/src/index.ts
slvrtrn: We do, but that's from the common package - `ClickHouseClient<Stream>`; I think it's better to re-export the platform-specific class as just `ClickHouseClient` instead and not expose `ClickHouseClient<Stream>` from the common at all. So both Node.js and Web packages will have the same name exported for the client class (currently, there is a slight discrepancy), and for its configuration params types, while in fact these will be just aliased exports.

mshustov (approved):

PR #242, opened 2024-03-13
Node.js connection idle sockets timeout rework, logger improvements

### LINKED ISSUE #196: Keep-Alive degrades performance
### Describe the bug
Using keep-alive with clickhouse-js degrades performance of concurrent queries compared to disabling keep-alive.

### Steps to reproduce
1. Clone [this](https://github.com/kiliangrashoff/clickhouse-js-performance) minimal reproduction example
2. Run it with the instructions in the readme
Output for me:
```
Average response time with keep alive true: 13.6ms
Average response time with keep alive false: 8.6ms
```

Numbers vary quite a bit, but keep alive false is consistently faster. Adding a `await result.json()` to the example improves performance of keep alive true, but this is not what we observe with real workloads.

With a real workload, performance of our api degrades from 95ms to 120ms average response time when enabling keep-alive. We configured keep_alive with the following settings and a keep_alive_timeout of 60s on the server, this does not fix the issue.
```
keep_alive: {
  enabled: true,
  socket_ttl: 55000,
  retry_on_expired_socket: true,
}
```

### Expected behaviour
Enabling keep-alive should result in better performance.

Executing a `SELECT 1` query twice with curl to [time the HTTP transaction](https://netbeez.net/blog/http-transaction-timing-breakdown-with-curl/) with an open TCP connection results in 0.6ms response time for the second request, I would expect the JS client to be able to query an open connection with similar response time:
```
Curl command:
curl --write-out 'lookup:        %{time_namelookup}\nconnect:       %{time_connect}\nappconnect:    %{time_appconnect}\npretransfer:   %{time_pretransfer}\nredirect:      %{time_redirect}\nstarttransfer: %{time_starttransfer}\ntotal:         %{time_total}\n'  localhost:8123/?query=SELECT%201 localhost:8123/?query=SELECT%201

Output:
1
lookup:        0.000031
connect:       0.000210
appconnect:    0.000000
pretransfer:   0.000278
redirect:      0.000000
starttransfer: 0.017353
total:         0.017902
1
lookup:        0.000009
connect:       0.000000
appconnect:    0.000000
pretransfer:   0.000023
redirect:      0.000000
starttransfer: 0.000445
total:         0.000553
```

### Code example
```
import { ClickHouseClient, createClient } from '@clickhouse/client'
import pLimit from 'p-limit'

const host = 'http://localhost:8123'
const username = 'default'
const password = undefined

const requests = 1000
const workers = 20

async function main() {
  await runPerformanceTest(true)
  await runPerformanceTest(false)
}

async function runPerformanceTest(keep_alive: boolean) {
  const client = await createClickHouseClient(keep_alive)
  const averageResponseTime = await measureAverageResponseTime(client, keep_alive)
  console.log(`Average response time with keep alive ${keep_alive}: ${averageResponseTime}ms`)
  await client.close()
}

async function measureAverageResponseTime(client: ClickHouseClient, keep_alive: boolean) {
  const limit = pLimit(workers)
  const queries = Array.from({length: requests}, () => limit(async () => {
    const start = process.hrtime()
    await client.query({
      query: 'SELECT 1',
      format: 'JSONEachRow'
    })
    const responseTime = process.hrtime(start)[1] / 1000000
    return responseTime
  }))
  const responseTimes = await Promise.all(queries)
  return responseTimes.reduce((a: number, b: number) => a + b, 0) / responseTimes.length
}

async function createClickHouseClient(keep_alive: boolean) {
  return createClient({
    host,
    username,
    password,
    keep_alive: {
      enabled: keep_alive
    },
    compression: {
      response: false,
      request: false
    }
  })
}

main().catch((err) => {
  console.error(err)
  process.exit(1)
})
```
Note: example depends on `p-limit` version `3.1.0` for concurrency control, easiest way to run it is to clone the repo I linked.

#### Environment
* Client version: 0.2.0
* NodeJS version: 18.17.0
* Typescript version: 5.1.6
* OS: Ubuntu 23.04

#### ClickHouse server
* ClickHouse Server version: 23.6.2.18

### LINKED ISSUE #237: Improve sockets housekeeping mechanism
Currently, implemented in a branch: https://github.com/ClickHouse/clickhouse-js/compare/main...idle-sockets-timeouts

It needs a rebase to 1.0.0.

General idea: instead of tracking if a socket _could be_ expired while idling and deciding what to do with it when making an HTTP request, just terminate it after (by default) 2500 ms using Node.js `Socket.on('free')` API.

slvrtrn:
## Summary
Initially, this PR was planned for 1.0.0, but it will be safer to release this as 0.3.0-beta.1 in advance to test this particular changeset.

Resolves #237 
Resolves #196 (it's not that KeepAlive was degrading performance; it was that KeepAlive=false was still using the entire KeepAlive mechanism).

Changes (some are breaking):
* Idle sockets timeout rework; now, the client tracks the idling sockets (on top of what the KeepAlive agent should normally do) and forcefully destroys them when it sees fit (by default, after 2500ms of idling, to match the default ClickHouse server KeepAlive configuration). 
* Ensure proper `Connection` header value considering KeepAlive settings. If KeepAlive is disabled, its value is now forced to ["close"](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection#close).
* (Breaking) `keep_alive.retry_on_expired_socket` and `keep_alive.socket_ttl` configuration parameters are removed. Instead, the new `keep_alive.idle_socket_ttl` parameter should be used. The default value for `keep_alive.idle_socket_ttl` is 2500.
* Fixed a bug with Ping that could lead to unhandled Socket hang-up propagation.
* Added an extra layer of "protection" against potential socket issues to the query/select/exec methods in the underlying HTTP connection implementation. If a request fails, the client will try to ensure that the outgoing request is also destroyed, meaning that the socket will be released immediately; this socket also shouldn't produce any unexpected "results" later (such as, again, unhandled Socket hang up from a rogue socket).
* (Technically breaking) The `max_open_connections` configuration parameter is now 10 by default, as we should not rely on the KeepAlive agent's defaults.
* (from 1.0.0, technically breaking) Backported the fix of the default `request_timeout` configuration value (now it is correctly set to 30s).
* Logger and logging improvements - a bit fancier default console writer and more logs on failing requests (if the log level is not `OFF` in the client configuration (it is off by default)); all client methods except ping will log an error on failure now, and ping will log a warning (cause the error is returned as part of its result).
* Split/fixed connection unit tests cause it already had too many test cases for a single file.
* Fixed a minor "lint" in `toURLSearchParams` since query_id has not been optional within the connection for a long time and is present for any request that uses it.
* From #234 - removed Node.js 16 from the CI and bumped GHA versions.

A changelog entry will be added if we decide to proceed with this.

## Checklist


.github/workflows/tests.yml
```diff
@@ -26,12 +26,12 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        node: [16, 18, 20]
+        node: [18, 20, 21]
```
mshustov, line 29: 21 is not LTS and might be unstable

packages/client-common/src/connection.ts
```diff
@@ -51,6 +51,8 @@ export type ConnPingResult =
     }
   | { success: false; error: Error }
 
+export type ConnOperation = 'Ping' | 'Query' | 'Insert' | 'Exec'
```
mshustov, line 54: nit: I'd rather use the fullname `ConnectorOperation` or just `Operation` to avoid any abbreviations

packages/client-common/src/logger.ts
```diff
@@ -4,78 +4,137 @@ export interface LogParams {
...
     })
   }
 
-  trace(params: LogParams): void {
+  trace(params: LogWriterParams<'trace'>): void {
     if (this.logLevel <= (ClickHouseLogLevel.TRACE as number)) {
-      this.logger.trace(params)
+      this.logger.trace({
+        ...params,
+        module: params.module ?? this.module,
```
mshustov, line 94: enhancement request: `context`-tree like in log4j

packages/client-node/__tests__/integration/node_socket_timeout.test.ts
```diff
@@ -0,0 +1,141 @@
...
+        enable: true,
+      },
+    } as NodeClickHouseClientConfigOptions)
+  })
+  afterAll(async () => {
+    await client.close()
+    server.close()
+  })
+
+  // ping first, then 2 operations in all possible combinations - repeat every combination several times
```
mshustov, line 40: wow that's something :D

packages/client-node/__tests__/integration/node_socket_timeout.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      const pingResult = await client.ping()
+      expect(pingResult.success).toBeFalse()
+      expect((pingResult as { error: Error }).error.message).toEqual(
+        jasmine.stringContaining('Timeout error.')
+      )
+    }
+    expect().nothing()
+  })
+
+  it('should not throw unhandled errors with Select', async () => {
```
mshustov, line 81: I don't think we test this. The test verifies `select` throws `Timeout error.` exception. How does just runner handles uncaught exceptions / promise rejections?
https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event
https://nodejs.org/api/process.html#event-unhandledrejection

slvrtrn: The test will still fail if something is thrown after the timeout error and it is unhandled (like something thrown from the Node's HTTP internals after we get the timeout error). This was the main problem with Ping, actually. That's also why there is a test of Ping + a few other operations in sequence.

packages/client-common/src/connection.ts
slvrtrn: I will rename all of these types in 1.0.0, then. They are all prefixed with just `Conn`.

.github/workflows/tests.yml
slvrtrn: It's worth keeping if it does not bring any flakiness to the CI run. If it starts to do so, we will remove it.

PR #245, opened 2024-03-15
Update examples

slvrtrn:
## Summary

- Added an overview with categories and a short description of what's inside each file. A link to this overview will be added to the docs, too.
- Improved async insert examples; previously, there was a very simple example with a vague name `insert_cloud`; now, it comes in two variations - with or without the `wait_for_async_insert` setting. Each variation is more extensive in detail.
- Removed "endless" stream examples as this, in my opinion, should be done with async inserts instead. `async_insert_without_waiting` targets the same use case as it was in the "endless" JSON stream example.
- More exhaustive `ping` usage overview (successful ping, non-existing host, timeout examples).
- Improved `abort_request`, `clickhouse_settings` examples
- Added `cancel_query` example.
- Added comments regarding supported data formats to the streaming examples.


examples/README.md
```diff
@@ -3,12 +3,74 @@
...
+
+Environment requirements for all examples:
+
+- Node.js 18+
+- NPM
+
+Run ClickHouse in Docker:
+
+```bash
+docker-compose up -d
```
mshustov, line 72: Run from the root of the repo? let's add it as a comment

examples/README.md
```diff
@@ -56,21 +118,22 @@ docker-compose -f docker-compose.cluster.yml up -d
...
 
 ```sh
 export CLICKHOUSE_HOST=https://<your-clickhouse-cloud-hostname>:8443
 export CLICKHOUSE_PASSWORD=<your-clickhouse-cloud-password>
 ```
 
-You can obtain these credentials in the Cloud console.
-This example assumes that you do not add any users or databases
-to your Cloud instance, so it is `default` for both.
+You can obtain these credentials in the ClickHouse Cloud console.
```
mshustov: should we add a link to the docs or a screenshot with the "Show connection string"?

examples/cancel_query.ts
```diff
@@ -0,0 +1,41 @@
...
+      if (err instanceof ClickHouseError && err.code === '394') {
+        console.error('Got an expected ClickHouse error:', err)
+      } else {
+        console.error('Unexpected error', err)
+      }
+    })
+
+  // Similarly, a mutation can be cancelled.
+  // See also: https://clickhouse.com/docs/en/sql-reference/statements/kill#kill-mutation
+  await client.command({
```
mshustov, line 31: dream: One day, to have the recipes on the website

examples/README.md
slvrtrn: There is a "Gather your connection details" section. https://clickhouse.com/docs/en/integrations/language-clients/javascript#gather-your-connection-details

examples/async_insert_without_waiting.ts
```diff
@@ -0,0 +1,126 @@
...
+    // Each individual insert operation will be resolved as soon as the request itself was processed on the server.
+    // The data will be batched on the server side. Insert will not wait for an ack about a successfully written batch.
+    // This is the main difference from the `examples/async_insert.ts` example.
+    await client
+      .insert({
+        table: tableName,
+        values: rows,
+        format: 'JSONEachRow',
+      })
+      .then(() => {
```
mshustov: why not use `await` syntax, IMO mixing both makes it harder to read

examples/async_insert_without_waiting.ts
```diff
@@ -0,0 +1,126 @@
...
+      ENGINE MergeTree()
+      ORDER BY (id)
+    `,
+  })
+
+  type Row = { id: number; name: string }
+
+  // Assume we have an event listener in our application that periodically receives incoming data,
+  // that we would like to have inserted into ClickHouse.
+  // This emitter is just a simulation for the sake of this example.
```
mshustov, line 44: the code can just call `void asyncInsertOnData(rows)`?

mshustov (approved):
great job with the examples :shipit:

examples/async_insert_without_waiting.ts
slvrtrn: You might be right. This emitter was written just for the sake of the example.

examples/README.md
slvrtrn: Added the link.

PR #247, opened 2024-03-18
More examples: select and insert formats overview

slvrtrn:
More examples: select and insert formats overview. 
I am not sure why we did not have this from the beginning, as these overviews are incredibly useful as a quick reference.

### DIFF
--- a/examples/insert_data_formats_overview.ts
+++ b/examples/insert_data_formats_overview.ts
@@ -1,0 +1,162 @@
+import { createClient } from '@clickhouse/client'
+import type { DataFormat } from '@clickhouse/client-common'
+import {
+  type InputJSON,
+  type InputJSONObjectEachRow,
+} from '@clickhouse/client-common'
+
+/**
+ * An overview of available formats for inserting your data, mainly in different JSON formats.
+ * For "raw" formats, such as:
+ *  - CSV
+ *  - CSVWithNames
+ *  - CSVWithNamesAndTypes
+ *  - TabSeparated
+ *  - TabSeparatedRaw
+ *  - TabSeparatedWithNames
+ *  - TabSeparatedWithNamesAndTypes
+ *  - CustomSeparated
+ *  - CustomSeparatedWithNames
+ *  - CustomSeparatedWithNamesAndTypes
+ *  - Parquet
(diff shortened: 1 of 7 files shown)


examples/insert_data_formats_overview.ts
```diff
@@ -0,0 +1,162 @@
+import { createClient } from '@clickhouse/client'
+import type { DataFormat } from '@clickhouse/client-common'
+import {
+  type InputJSON,
+  type InputJSONObjectEachRow,
+} from '@clickhouse/client-common'
+
+/**
+ * An overview of available formats for inserting your data, mainly in different JSON formats.
+ * For "raw" formats, such as:
```
mshustov, line 10: Let's add a comment: The list might be incomplete, so we ask users to submit more examples. It's too easy to add a new format without updating the example.

examples/insert_data_formats_overview.ts
```diff
@@ -0,0 +1,162 @@
...
+  console.log('\n#### Single document JSON formats:\n')
+  // JSON, JSONCompact, JSONColumnsWithMetadata accept the InputJSON<T> shape.
+  // For example: https://clickhouse.com/docs/en/interfaces/formats#json
+  const meta: InputJSON['meta'] = [
+    { name: 'id', type: 'UInt32' },
+    { name: 'name', type: 'String' },
+    { name: 'sku', type: 'Array(UInt32)' },
+  ]
+  await insertJSON('JSON', {
+    meta: [], // not required for JSON format input
```
mshustov, line 90: Let's remove it then?

examples/package.json
```diff
@@ -12,7 +12,10 @@
   },
   "scripts": {},
   "dependencies": {
-    "@clickhouse/client": "latest",
+    "@clickhouse/client": "latest"
+  },
+  "devDependencies": {
+    "split2": "^4.2.0",
```
mshustov, line 18: 😄 did we missed it before?

mshustov (approved):

examples/package.json
slvrtrn: It could be - my local cache might have had it. Or maybe it comes from the dev dependencies of the client itself.

Issue #244, opened 2024-03-14
What is the best way to trigger a command and polling until is done by query_id?


francescorivola:
Hi ClickHouse Js Team,

I am writing to ask you what is your recommendation to achieve the following use case:

- We want to run a command that insert into a table the result of a complex query.

This command takes a quite large amount of time (maybe several minutes or more). We do not want to keep the client waiting for the response, instead just take the query_id and periodically check for system.query_log and check when the query is completed.

A possible solution could set a small request_timeout setting but that will likely require to create an instance of the Client for each query done for this process and I not sure is a good approach (However something similar we are doing right now in production with the lib https://github.com/TimonKK/clickhouse). The number of these operations in production are around 2K commands every hour.

Thank you in advance for your support and guidance.

slvrtrn:
If it's a simple INSERT INTO query with an array, then maybe you could try [async inserts](https://clickhouse.com/docs/en/optimize/asynchronous-inserts) (with disabled waiting on the client side).

```ts
const client = createClient({
  clickhouse_settings: {
    async_insert: 1,
    wait_for_async_insert: 0, // <- don't wait for ACK
  },
})

await client.insert({
  table,
  format: 'JSONEachRow', // or other, depending on your data
  values: [{ ... }],
})
```

francescorivola:
Hi @slvrtrn , the use case is an **INSERT INTO SELECT**. So we use the query result between other tables to populate a target table. Do you need more info?

slvrtrn:
It looks like that suspending an `INSERT FROM SELECT` statement might not be currently possible, see [this](https://github.com/ClickHouse/ClickHouse/issues/49683) issue in the main repo.

I'd be curious to learn more about your use case, cause maybe [materialized views](https://clickhouse.com/docs/en/guides/developer/cascading-materialized-views) will do.

francescorivola:
The use case is quite tricky, probably not the most common use case for ClickHouse users hehehe. Unfortunately I cannot enter too much in detail, I am sorry. 
But in the end we basically use ClickHouse to compute a dynamic result set (our customer can define dynamic filters and queries), this result is then stored in a table in ClickHouse to then be consumed as a stream by our transactional database. We store the result in a ClickHouse table instead of consuming directly the query result to allow the process to resume in case of an application graceful shutdown, the result set is stored in a proper order and can be consumed at any time resuming by the last processed row.

[...]

PR #251, opened 2024-03-28
Add pathname config option, revert readonly switch/default settings

slvrtrn:
## Summary

Into 1.0.0.

* Added `pathname` config option, as after https://github.com/ClickHouse/clickhouse-js/pull/232 we could have #207 issue again.
* Base and URL configs are now "deep merged" instead of overriding only the top-level key.
* Reverted #225: as the HTTP module has a header size limit (in my tests, around 76 progress headers), it is unsafe to enable it by default. Documentation and examples should already cover these long-running query use cases.
* Disabled `compression.response` by default; it was unclear what client option was causing issues with queries performed by the readonly=1 user when using the client without additional configuration.
* Added more advanced example on possible "hacks" with long-running queries based on our discussion in https://github.com/ClickHouse/clickhouse-js/issues/244
* Updated JSDoc for the config options.
* Adjusted changelog accordingly.

## Checklist


mshustov (approved):
The next time, I'd suggest splitting unrelated changes into 2 separate PRs to simplify the code review

PR #253, opened 2024-03-29
Improved performance when decoding the entire set of rows with streamable JSON formats

slvrtrn:
## Summary

* Improved performance when decoding the entire set of rows with _streamable_ JSON formats (such as `JSONEachRow` or `JSONCompactEachRow`) by calling the `ResultSet.json()` method. Depending on the dataset, it's between 10-15% (like [cell_towers](https://clickhouse.com/docs/en/getting-started/example-datasets/cell-towers)) and 40% (with large rows and very long strings) less execution time. NB: The actual streaming performance when consuming the `ResultSet.stream()` (which was fast) hasn't changed. Only the `ResultSet.json()` method used a suboptimal stream processing in some instances, and now `ResultSet.json()` just consumes the same stream transformer provided by the `ResultSet.stream()` method.

  Before:
  ```
  [1277 ms   ][JSONEachRow       ][SELECT * FROM large_strings ORDER BY id ASC LIMIT 100000                        ]
  [668 ms    ][JSONEachRow       ][SELECT * FROM cell_towers ORDER BY (radio, mcc, net, created) ASC LIMIT 500000  ]
  ```
  After:
  ```
  [737 ms    ][JSONEachRow       ][SELECT * FROM large_strings ORDER BY id ASC LIMIT 100000                        ]
  [578 ms    ][JSONEachRow       ][SELECT * FROM cell_towers ORDER BY (radio, mcc, net, created) ASC LIMIT 500000  ]
  ```

* Removed the outdated `decode` function.
* Updated exported types and doc entries for DataFormat.
* Fixed weird flakiness in expect-type assertions

### DIFF
--- a/packages/client-web/__tests__/integration/web_select_streaming.test.ts
+++ b/packages/client-web/__tests__/integration/web_select_streaming.test.ts
@@ -40,9 +40,9 @@ describe('[Web] SELECT streaming', () => {
     it('should consume a text response only once', async () => {
       const rs = await client.query({
         query: 'SELECT * FROM system.numbers LIMIT 1',
-        format: 'TabSeparated',
+        format: 'JSONEachRow',
       })
-      expect(await rs.text()).toEqual('0\n')
+      expect(await rs.text()).toEqual('{"number":"0"}\n')
       // wrap in a func to avoid changing inner "this"
       await assertAlreadyConsumed$(() => rs.json())
       await assertAlreadyConsumed$(() => rs.text())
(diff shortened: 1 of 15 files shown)


packages/client-web/__tests__/integration/web_select_streaming.test.ts
```diff
@@ -40,9 +40,9 @@ describe('[Web] SELECT streaming', () => {
     it('should consume a text response only once', async () => {
       const rs = await client.query({
         query: 'SELECT * FROM system.numbers LIMIT 1',
-        format: 'TabSeparated',
+        format: 'JSONEachRow',
```
mshustov, line 43: why do we need the change?

mshustov (approved):
What change gave the performance boost? I see changes in `restul_set` and `stream`

packages/client-web/__tests__/integration/web_select_streaming.test.ts
slvrtrn: can't call JSON on TabSeparated.

slvrtrn:
@mshustov 
> What change gave the performance boost? 
The `decode` function did a few unneeded passes on a huge string. It is way less efficient than the stream transformer we already had in the ResultSet.

slvrtrn:
This: https://github.com/ClickHouse/clickhouse-js/pull/253/files#diff-46e9eda9880b41c04095d911536817d628992924b2bdd3a96f66529dd3628a5bL101-L104

Unless it's optimized out by JIT, it's at least one extra full pass, worst case two.

PR #264, opened 2024-05-07
Drain the response stream in the command method

slvrtrn:
## Summary

The `command` method now drains the response stream properly, as the previous implementation could cause the Keep-Alive socket to close after each request.

## Checklist

### DIFF
--- a/packages/client-node/src/connection/node_base_connection.ts
+++ b/packages/client-node/src/connection/node_base_connection.ts
@@ -1,6 +1,7 @@
 import type {
   ClickHouseSummary,
   ConnBaseQueryParams,
+  ConnCommandResult,
   Connection,
   ConnectionParams,
   ConnExecResult,
@@ -27,7 +28,6 @@ import Zlib from 'zlib'
 import { getAsText, getUserAgent, isStream } from '../utils'
 import { decompressResponse, isDecompressionError } from './compression'
 import { drainStream } from './stream'
-import type { ConnCommandResult } from '@clickhouse/client-common'
 
 export type NodeConnectionParams = ConnectionParams & {
   tls?: TLSParams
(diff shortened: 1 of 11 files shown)


packages/client-node/src/connection/node_base_connection.ts
```diff
@@ -80,204 +81,6 @@ export abstract class NodeBaseConnection
...
-      request.on('close', onClose)
-
-      if (params.abort_signal !== undefined) {
-        params.abort_signal.addEventListener('abort', onAbort, { once: true })
-      }
-
-      if (!params.body) return request.end()
-    })
-  }
-
```
slvrtrn, line 280: The `buildDefaultHeaders`/`createClientRequest`/`request` methods were moved to the bottom cause it was too annoying to have it at the very top (no other changes).

packages/client-node/src/connection/node_base_connection.ts
```diff
@@ -358,52 +161,6 @@ export abstract class NodeBaseConnection
...
-        err: err as Error,
-        extra_args: {
-          clickhouse_settings: params.clickhouse_settings ?? {},
-        },
-      })
-      throw err // should be propagated to the user
-    } finally {
-      controllerCleanup()
-    }
-  }
```
slvrtrn, line 405: This one was used for both exec + command client methods. Now, the connection has them separate, but the logic is still (mostly) shared.

packages/client-web/src/config.ts
```diff
@@ -18,5 +18,4 @@ export const WebImpl: ImplementationDetails<ReadableStream>['impl'] = {
     query_id: string,
   ) => new ResultSet(stream, format, query_id)) as any,
   values_encoder: new WebValuesEncoder(),
-  close_stream: (stream) => stream.cancel(),
```
slvrtrn, line 21: Finally, we don't need this anymore.

mshustov (approved):

Issue #271, opened 2024-05-22
BaseQueryParams.session_id is not used


Holi0317:
### Describe the bug

`BaseQueryParams` got an [undocumented `session_id`](https://github.com/ClickHouse/clickhouse-js/blob/ad8611e736a82d98416c1d99d903240df02d7ace/packages/client-common/src/client.ts#L28) that feels like it should override the `session_id` passed into `createClient`. But actually that field is never used and `session_id` is always using `this.sessionId` https://github.com/ClickHouse/clickhouse-js/blob/ad8611e736a82d98416c1d99d903240df02d7ace/packages/client-common/src/client.ts#L253

### Steps to reproduce
1. Try to override `session_id` via `BaseQueryParams`

### Expected behaviour

The `session_id` should be overridden

slvrtrn:
Added in [1.1.0](https://github.com/ClickHouse/clickhouse-js/releases/tag/1.1.0).

PR #272, opened 2024-05-22
Allow override session_id in exec

Holi0317:
## Summary

Allow overriding `session_id` per query

https://github.com/ClickHouse/clickhouse-js/issues/271

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# UNRELEASED (Common, Node.js, Web)
+
+## Bug fixes
+
+- Allow overriding `session_id` per query ([@holi0317](https://github.com/Holi0317), [#271](https://github.com/ClickHouse/clickhouse-js/issues/271)).
+
 # 1.0.2 (Common, Node.js, Web)
 
 ## Bug fixes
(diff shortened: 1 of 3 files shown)


slvrtrn:
Thanks for your contribution! Nice catch.

A few notes:
- the `exec` method is not intended to be used with `CREATE TABLE` statements; the relevant method is `command`. The main purpose of `exec` is to provide a workaround for "unusual" statements that will provide a useful output ([select with a default format](https://github.com/ClickHouse/clickhouse-js/blob/50f790f660dcb9d91e7c224c411dd8cbbe7055b2/examples/default_format_setting.ts#L12), [select into a file](https://github.com/ClickHouse/clickhouse-js/blob/b740afe9c2a238837c339264ad18ae645b69a5a6/examples/node/select_parquet_as_file.ts#L15)) and something similar; if some of the docs/examples are misleading, let's update that as well.
- that said, it is worth checking if we can do a similar override in the "main" three methods - `query`, `command`, `insert`; I am AFK now, so can't check it quickly.
- If it is indeed the case that we cannot override the session id in these methods, let's add it there, too, and enhance the integration test to use all three methods (command to create the table, insert to insert the data, query to select). Maybe the same for `exec`.
- Note: If you use `exec` to create the table, don't forget to destroy the returned stream. This is done internally in the `command` method (though the stream is not destroyed, but "drained" instead, to preserve the keep-alive socket).

slvrtrn:
I will recreate it as a new PR from another branch (where I have write access).

PR #279, opened 2024-05-28
Allow credentials override for a particular request

### LINKED ISSUE #278: Request for Connection Pooling Feature when creating several clients.
Hey,

We are using the Clickhouse node package to create queries from an app to our DB. 
We have many users who frequently query the database, Each user has a unique Clickhouse user, requiring the creation of a new client object for each query. This results in significant TCP connection overhead.

We would love the option to have a connection pool to prevent the TCP overhead associated with creating new client objects. For example, adding the option to send an agent to the client, similar to Axios, to get the connection from a pool instead of creating a new TCP connection. This would allow us to reuse existing connections, optimizing performance and reducing latency.

Thank you very much for looking into this request, looking forward to hearing back from you.

Best regards,
Yosef, TripleWhale.

slvrtrn:
## Summary

Allows to override username/password for a particular request (query/command/exec/insert).
Resolves #278 

## Checklist

### DIFF
--- a/packages/client-common/__tests__/integration/auth.test.ts
+++ b/packages/client-common/__tests__/integration/auth.test.ts
@@ -29,7 +29,7 @@ describe('authentication', () => {
     )
   })
 
-  describe('auth override', () => {
+  describe('request auth override', () => {
     let defaultClient: ClickHouseClient
     beforeAll(() => {
       defaultClient = createTestClient()
--- a/packages/client-node/__tests__/tls/tls.test.ts
+++ b/packages/client-node/__tests__/tls/tls.test.ts
@@ -20,7 +20,7 @@ describe('[Node.js] TLS connection', () => {
 
   it('should work with basic TLS', async () => {
     client = createClient({
-      host: 'https://server.clickhouseconnect.test:8443',
+      url: 'https://server.clickhouseconnect.test:8443',
       tls: {
         ca_cert,
       },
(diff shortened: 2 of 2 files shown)


packages/client-node/__tests__/utils/http_stubs.ts
```diff
@@ -134,6 +134,9 @@ export class MyTestHttpConnection extends NodeBaseConnection {
     return {} as any
   }
   public getDefaultHeaders() {
-    return this.headers
```
mshustov, line 137: Should we add an integration test?

packages/client-common/src/client.ts
```diff
@@ -23,9 +24,16 @@ export interface BaseQueryParams {
...
   abort_signal?: AbortSignal
   /** A specific `query_id` that will be sent with this request.
-   * If it is not set, a random identifier will be generated automatically by the client. */
+   *  If it is not set, a random identifier will be generated automatically by the client. */
   query_id?: string
   session_id?: string
+  /** When defined, overrides the username from the {@link BaseClickHouseClientConfigOptions.username} setting.
+   *  @default undefined */
+  username?: string
+  /** Has no effect if the {@link BaseQueryParams.username} value is not set.
```
mshustov: considering that we require both `username` and `password` to be either defined or not, maybe we should put it under a single field
```typescript
auth?: { username: string, password: string } | { token: string }
```

mshustov (approved):

Issue #283, opened 2024-06-15
Allow support for either multiple queries or more direct control of the agent to ensure queries are sent via the same socket session


brian-mann:
### Use case
We need the ability to issue multiple requests using the same socket connection in order to utilize creating temporary tables and querying them.

The `clickhouse cli` binary allows for the `--multiquery` flag but there is no way to issue multiple queries (to my knowledge) with the nodejs client.

For instance it does not seem easily possible to do this:

```sql
create temporary table foo as select 1;
select * from foo;
```

The only way we've found is to make 2 round trips, but to do so, we have to ensure we're using the same socket connection. An alternative approach would be to more directly control the underlying node.js agent.

There was a PR opened here (https://github.com/ClickHouse/clickhouse-js/pull/275) enabling passing in your own agent instance, but the PR was closed.

### Describe the solution you'd like

We're also aware of the feature request to materialize CTE's (PR opened here: https://github.com/ClickHouse/ClickHouse/pull/61086) - which would also solve for our use case, but until that is implemented - either having multiquery support OR making it easier to control the request agent is the only solution.

More direct control of the agent would at least give us the ability to use the same connection pool, but ensure that the socket connection is not free'd up until the dependent queries have finished. Ideally though, multiquery or materialized CTE's would be a superior solution.

### Describe the alternatives you've considered

As a workaround, we can effectively recreate this by instantiating a new client and setting `max_open_connections: 1` so that all queries run over the same socket, but that is not only very inefficient because the socket has to be reconnected and closed each time, but it bypasses the natural connection pool.

mshustov:
>  but ensure that the socket connection is not free'd up until the dependent queries have finished.

I don't think it's safe to rely on this assumption in distributed enviroment with network hiccups. 

> More direct control of the agent would at least give us the ability to use the same connection pool, but ensure that the socket connection is not free'd up until the dependent queries have finished. 

How do you deploy ClickHouse? Is there an option of using a session mechanism? from the [temporary table docs](https://clickhouse.com/docs/en/sql-reference/statements/create/table#temporary-tables)
> primary use case where temporary tables can be useful is for querying or joining small external datasets during a single session.

slvrtrn:
@brian-mann, it will be possible to override the HTTP(s) agent after https://github.com/ClickHouse/clickhouse-js/pull/284.

[...]

PR #284, opened 2024-06-18
Allow to provide a custom http agent and disable the auth header

slvrtrn:
## New features

- Added an option to provide a custom HTTP Agent in the client configuration via the `http_agent` option ([#283](https://github.com/ClickHouse/clickhouse-js/issues/283), related: [#278](https://github.com/ClickHouse/clickhouse-js/issues/278)). The following conditions apply if a custom HTTP Agent is provided:
  - The `max_open_connections` and `tls` options will have _no effect_ and will be ignored by the client, as it is a part of the underlying (default) HTTP Agent configuration.
  - `keep_alive.enabled` will only regulate the default value of the `Connection` header (`true` -> `Connection: keep-alive`, `false` -> `Connection: close`).
  - While the idle socket management will still work, it is now possible to disable it completely by setting the `keep_alive.idle_socket_ttl` value to `0`.
- Added a new client configuration option: `set_basic_auth_header` which disables the `Authorization` header that is set by the client by default for every outgoing HTTP request. One of the possible scenarios when it is necessary to disable this header is when a custom HTTPS agent is used, and the server requires TLS authorization with certificates. For example:

  ```ts
  const agent = new https.Agent({
    ca: fs.readFileSync('./ca.crt'),
  })
  const client = createClient({
    url: 'https://server.clickhouseconnect.test:8443',
    http_agent: agent,
    // With a custom HTTPS agent, the client won't use the default HTTPS connection implementation; the basic TLS headers should be provided manually
    http_headers: {
      'X-ClickHouse-User': 'default',
      'X-ClickHouse-Key': '',
    },
    // Authorization header conflicts with the TLS headers; disable it.
    set_basic_auth_header: false,
  })
  ```

NB: It is currently not possible to set the `set_basic_auth_header` option via the URL params.

## Checklist
- [x] For significant changes, documentation in https://github.com/ClickHouse/clickhouse-docs was updated with further explanations or tutorials - https://github.com/ClickHouse/clickhouse-docs/pull/2412

### DIFF
--- a/packages/client-node/__tests__/tls/tls.test.ts
+++ b/packages/client-node/__tests__/tls/tls.test.ts
@@ -1,6 +1,8 @@
 import type { ClickHouseClient } from '@clickhouse/client-common'
 import { createTestClient } from '@test/utils'
 import * as fs from 'fs'
+import Http from 'http'
+import https from 'node:https'
 import type Stream from 'stream'
 import { createClient } from '../../src'
 
@@ -134,6 +136,37 @@ describe('[Node.js] TLS connection', () => {
         },
       })
       expect(await resultSet.text()).toEqual('0\n1\n2\n')
+    })
+
+    describe('Custom HTTPS agent', () => {
+      let httpRequestStub: jasmine.Spy<typeof Http.request>
+      beforeEach(() => {
+        httpRequestStub = spyOn(Http, 'request').and.callThrough()
+      })
+
+      it('should work with a custom HTTPS agent', async () => {
+        const agent = new https.Agent({
+          maxFreeSockets: 5,
+          ca: ca_cert,
+        })
+        const client = createClient({
+          url: 'https://server.clickhouseconnect.test:8443',
+          http_agent: agent,
+          http_headers: {
+            'X-ClickHouse-User': 'default',
+            'X-ClickHouse-Key': '',
+          },
+          set_basic_auth_header: false,
+        })
+        const rs = await client.query({
+          query: 'SELECT 144 AS result',
+          format: 'JSONEachRow',
+        })
+        expect(await rs.json()).toEqual([{ result: 144 }])
+        expect(httpRequestStub).toHaveBeenCalledTimes(1)
+        const callArgs = httpRequestStub.calls.mostRecent().args
+        expect(callArgs[1].agent).toBe(agent)
+      })
     })
   })
 })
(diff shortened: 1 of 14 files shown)


slvrtrn:
Docs: https://github.com/ClickHouse/clickhouse-docs/pull/2412

packages/client-node/__tests__/tls/tls.test.ts
```diff
@@ -135,5 +137,36 @@ describe('[Node.js] TLS connection', () => {
...
+        httpRequestStub = spyOn(Http, 'request').and.callThrough()
+      })
+
+      it('should work with a custom HTTPS agent', async () => {
+        const agent = new https.Agent({
+          maxFreeSockets: 5,
+          ca: ca_cert,
+        })
+        const client = createClient({
+          url: 'https://server.clickhouseconnect.test:8443',
```
mshustov, line 153: can't we run the test against Cloud?

packages/client-node/src/config.ts
```diff
@@ -22,10 +24,21 @@ export type NodeClickHouseClientConfigOptions =
...
+    /** Custom HTTP agent to use for the outgoing HTTP(s) requests.
+     *  If set, {@link BaseClickHouseClientConfigOptions.max_open_connections}, {@link tls} and {@link keep_alive}
+     *  options have no effect, as it is part of the default underlying agent configuration.
+     *  @experimental - unstable API, might be a subject to change in the future; please provide your feedback in the repository.
+     *  @default undefined */
+    http_agent?: http.Agent | https.Agent
+    /** Enable or disable the `Authorization` header with basic auth for the outgoing HTTP(s) requests.
+     *  @experimental - unstable API, might be a subject to change in the future; please provide your feedback in the repository.
+     *  @default true (enabled) */
+    set_basic_auth_header?: boolean
```
mshustov, line 41: what if we change the client signature
```typescript
  /** The name of the user on whose behalf requests are made.
   *  @default default */
  username?: string | null
  /** The user password.
   *  @default empty string */
  password?: string | null
```

if `null` is provided, no defaults are applied, and no auth header is set?

mshustov (approved):

packages/client-node/__tests__/tls/tls.test.ts
slvrtrn: We can, but this is slightly more annoying - it will have to choose between HTTP and HTTPS agent, plus I wanted to test the auth with a CA cert (so I'll need to add that to the repo as well).

PR #288, opened 2024-07-05
Expose full response headers object

slvrtrn:
## Summary

It is now possible to get the entire response headers object from the `query`/`insert`/`command`/`exec` methods. With `query`, you can access the `ResultSet.response_headers` property; other methods (`insert`/`command`/`exec`) return it as parts of their response objects as well.

  For example:

  ```ts
    query: 'SELECT * FROM system.numbers LIMIT 1',
    format: 'JSONEachRow',
  })
  console.log(rs.response_headers['content-type'])
  ```

  This will print: `application/x-ndjson; charset=UTF-8`. It can be used in a similar way with the other methods.

## Checklist


fneves (approved):

packages/client-web/src/connection/web_connection.ts
```diff
@@ -69,15 +71,16 @@ export class WebConnection implements Connection<ReadableStream> {
     return {
       query_id: result.query_id,
       stream: result.stream || new ReadableStream<Uint8Array>(),
+      response_headers: result.response_headers,
```
mshustov, line 74: It's the first HTTP abstraction that leaks in the result set. I'd rather add some generic method to avoid a breaking change once TCP protocol is supported, but your call

mshustov (approved):

packages/client-web/src/connection/web_connection.ts
slvrtrn: I think it is already leaking a bit, as we allow `text`/`json` calls, which are not very compatible with `Native` and even `RowBinary` format; in time, we will likely have to implement the native version with a new result set flavor (similarly to `RowBinary`).

PR #290, opened 2024-07-11
An optional `values` parameter in the `exec` method

slvrtrn:
## Summary

The `exec` method now accepts an optional `values` parameter, which allows you to pass the request body as a `Stream.Readable`. This can be useful in case of custom insert streaming with arbitrary ClickHouse data formats (which might not be explicitly supported and allowed by the client in the `insert` method yet). 

A very minimal example with CSV (from the tests, it does not make sense as CSV is supported via the normal `insert`, just an illustration):

```ts
const stream = Stream.Readable.from(['42,foobar,"[1,2]"'], {
  objectMode: false,
})
const execResult = await client.exec({
  query: `INSERT INTO ${tableName} FORMAT CSV`,
  values: stream,
})
// the result stream contains nothing useful for an insert 
// and should be immediately drained to release the socket
await drainStream(execResult.stream)

  query: `SELECT * FROM ${tableName}`,
  format: 'JSONEachRow',
})
expect(await rs.json()).toEqual([
  {
    id: '42',
    name: 'foobar',
    sku: [1, 2],
  },
])
```

## Checklist

### DIFF
--- a/packages/client-common/src/client.ts
+++ b/packages/client-common/src/client.ts
@@ -61,9 +61,16 @@ export type QueryResult<Stream, Format extends DataFormat> =
     ? BaseResultSet<Stream, unknown>
     : BaseResultSet<Stream, Format>
 
-export interface ExecParams extends BaseQueryParams {
-  /** Statement to execute. */
+export type ExecParams = BaseQueryParams & {
+  /** Statement to execute. By default, the query will be sent in the request body;
+   *  If {@link ExecParamsWithValues.values} are defined, the query is sent as a request parameter,
+   *  and the values are sent in the request body instead. */
   query: string
+}
+export type ExecParamsWithValues<Stream> = ExecParams & {
+  /** If you have a custom INSERT statement to run with `exec`,
+   *  the data from this stream will be inserted. */
+  values: Stream
 }
 
 export type CommandParams = ExecParams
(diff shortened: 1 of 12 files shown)


packages/client-common/src/client.ts
```diff
@@ -214,10 +226,14 @@ export class ClickHouseClient<Stream = unknown> {
    * but format clause is not applicable. The caller of this method is expected to consume the stream,
    * otherwise, the request will eventually be timed out.
    */
-  async exec(params: ExecParams): Promise<ExecResult<Stream>> {
+  async exec(
+    params: ExecParams | ExecParamsWithValues<Stream>,
```
slvrtrn, line 230: I avoided parametrizing ExecParams in this case, as it's a bit tricky to maintain full backward compatibility.

packages/client-node/src/connection/node_base_connection.ts
```diff
@@ -368,20 +369,23 @@ export abstract class NodeBaseConnection
...
+    }
+    const searchParams = toSearchParams(toSearchParamsOptions)
     const { controller, controllerCleanup } = this.getAbortController(params)
     try {
       const { stream, summary, response_headers } = await this.request(
         {
           method: 'POST',
           url: transformUrl({ url: this.params.url, searchParams }),
-          body: params.query,
+          body: sendQueryInParams ? params.values : params.query,
```
mshustov, line 388: nit: I'd prefer to keep a single logic path to send a request, but let's postpone until the next major release when we can *always* send `query` in URL and `values` as body (even if it's empty)

packages/client-node/__tests__/integration/node_exec.test.ts
```diff
@@ -66,4 +68,33 @@ describe('[Node.js] exec result streaming', () => {
       expect(await rs.json()).toEqual([{ number: '0' }])
     })
   })
+
+  describe('custom insert streaming with exec', () => {
```
mshustov, line 72: can we add a test that it doesn't blow with an empty stream, closed stream, and other edge cases?

mshustov (approved):

PR #301, opened 2024-08-20
Bump Node.js version, add wake-up-cloud step

slvrtrn:


.github/workflows/tests.yml
```diff
@@ -192,13 +192,28 @@ jobs:
         run: |
           npm run test:web:integration:local_cluster
 
+  wake-up-cloud:
+    needs:
+      - node-unit-tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@main
+      - name: Wake up the cloud instance
```
mshustov, line 201: I'm wondering what problem we want to address? Idling is disabled on `GitHub actions` instance

slvrtrn: I see. Then, something else happens during the test runs that causes an occasional timeout.

Issue #273, opened 2024-05-22
More specific types for `meta.type`


tylersayshi:
### Use case

Use case will be for handling formatting of the data value from a response, when `meta` is present on an app.

### Describe the solution you'd like

https://github.com/ClickHouse/clickhouse-js/blob/main/packages/client-common/src/clickhouse_types.ts#L8

This line ^ could be something more like the following

```ts
/** @see https://clickhouse.com/docs/en/sql-reference/data-types */
const DATA_TYPES = [
    'UInt8', 'UInt16', 'UInt32', 'UInt64', 'UInt128', 'UInt256', 'Int8', 'Int16', 'Int32', 'Int64', 'Int128', 'Int256',
    'Float32', 'Float64',
    'Boolean', 
    'String', 'FixedString',
    'Date', 'Date32',
    'DateTime', 'DateTime64',
    'JSON',
    'UUID',
    'Enum', 'LowCardinality',
    'Array',
    'Map',
    'SimpleAggregateFunction', 'AggregateFunction',
    'Nested',
    'Tuple',
    'Nullable',
    'IPv4', 'IPv6',
    'Point', 'Ring', 'Polygon', 'MultiPolygon', 
    'Expression', 'Set', 'Nothing', 'Interval'
] as const;
type DataType = typeof DATA_TYPES[number];

// ...
    meta?: Array<{
        name: string;
        type: DataType;
    }>;
```

### Describe the alternatives you've considered

This type could be an enum or string literal. I'd have no real preference. 

### Additional context

What would be nice for me as a consumer of `@clickhouse/client` is to be able to treat the library as a comprehensive source of truth for what data types I should expect to be able to format. This way, I can format them accordingly in my apps (especially if new data types are added later on or renamed, so I'd see the type change).

slvrtrn:
Array, Tuple, Map, DateTime, etc, are a bit tricky in this case cause:

```
➜  ~ curl  --data-binary \
     "SELECT tuple(32, 'foo') AS x FORMAT JSON;" \
     http://localhost:8123
{
	"meta":
	[
		{
			"name": "x",
			"type": "Tuple(UInt8, String)"
		}
	],

	"data":
	[
		{
			"x": [32,"foo"]
		}
	],

	"rows": 1,

	"statistics":
	{
		"elapsed": 0.000961292,
		"rows_read": 1,
		"bytes_read": 1
	}
}
```

So, the proposed literal won't match the actual type in certain cases.

[...]

PR #257, opened 2024-04-02
WIP: RowBinary for SELECT operations

slvrtrn:
## Summary

RowBinary format implementation for selects (resolves #216).

## Checklist
- [ ] Unit and integration tests covering the common scenarios were added
- [ ] A human-readable description of the changes was provided to include in CHANGELOG
- [ ] For significant changes, documentation in https://github.com/ClickHouse/clickhouse-docs was updated with further explanations or tutorials

PR #316, opened 2024-09-18
Expose `parseColumnType` function

slvrtrn:
## Summary

Extracted and exported the `parseColumnType` method from the #257 branch. Related to #273.
This is useful not only for RowBinary/Native protocols but also for TS interface codegens, so it is probably worth adding it right away.

For example:
- Input type: `Array(Array(DateTime('Etc/GMT-5')))`
- Output object:
```ts
{
  sourceType: `Array(Array(DateTime('Etc/GMT-5')))`,
  type: 'Array',
  dimensions: 2,
  value: {
    type: 'DateTime',
    timezone: 'Etc/GMT-5',
    sourceType: `DateTime('Etc/GMT-5')`,
  },
}
```

Currently supported types to parse:

* `Bool`
* `UInt8`
* `Int8`
* `UInt16`
* `Int16`
* `UInt32`
* `Int32`
* `UInt64`
* `Int64`
* `UInt128`
* `Int128`
* `UInt256`
* `Int256`
* `Float32`
* `Float64`
* `String`
* `FixedString(N)`
* `UUID`
* `Date`
* `Date32`
* `IPv4`
* `IPv6`
* `Array` (arbitrary nesting)
* `Map` (arbitrary nesting)
* `Tuple` (arbitrary nesting)
* `Enum` (including enums in arrays, maps, tuples)
* `DateTime` (including timezones)
* `DateTime64` (including timezones and various precisions)
* `Decimal`
* `Nullable`, `LowCardinality` combinations

Currently, not supported (will throw):

* `Geo`
* `(Simple)AggregateFunction`
* `Nested`
* Old/new `JSON`
* `Dynamic`
* `Variant`

Please take a look at the tests for more examples.
The tests are not fully complete, so this will be somewhat "experimental" still.

## Checklist


packages/client-node/src/index.ts
```diff
@@ -46,4 +46,17 @@ export {
...
+  type ParsedColumnFixedString,
+  type ParsedColumnNullable,
+  type ParsedColumnDecimal,
+  type ParsedColumnDateTime,
+  type ParsedColumnDateTime64,
+  type ParsedColumnArray,
+  type ParsedColumnTuple,
+  type ParsedColumnMap,
+  type ParsedColumnType,
+  parseColumnType,
```
mshustov, line 61: Neither nodejs nor web clients use any of these by default. Should we consider releasing it as a separate package? 
What benefits does the current distribution model have? I see a few disadvantages: larger distro size and harder to explore the original source code

packages/client-common/__tests__/unit/parse_column_types_enum.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    })
+  })
+
+  it('should throw when Enum has no values to parse', async () => {
+    // The minimal allowed Enum definition is Enum8('' = 0), i.e. 6 chars inside.
+    const allEnumTypeArgs: string[][] = enumTypes.map(([enumType]) => [
+      `${enumType}()`,
+      `${enumType}(')`,
+      `${enumType}('')`,
+      `${enumType}('' )`,
```
mshustov, line 77: ~could you add a test that an empty string is a valid value?~ found it below

mshustov (approved):

packages/client-node/src/index.ts
slvrtrn: > Neither nodejs nor web clients use any of these by default. Should we consider releasing it as a separate package?

If we are ever to add Native or RowBinaryWithNamesAndTypes, then this parser will be required anyway. Or do you mean that it should not be exported from the Node.js client, only in common?

> harder to explore the original source code

WDYM?

PR #328, opened 2024-09-25
Add support for role query parameters

### LINKED ISSUE #269: Add role support via query parameters
See https://github.com/ClickHouse/ClickHouse/pull/62669 and https://clickhouse.com/docs/en/interfaces/http#setting-role-with-query-parameters

pulpdrew:
## Summary

This PR adds support for specifying roles via request query parameters. Closes #269.

## Checklist

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -67,6 +67,7 @@ If something is missing, or you found a mistake in one of these examples, please
 - [default_format_setting.ts](default_format_setting.ts) - sending queries using `exec` method without a `FORMAT` clause; the default format will be set from the client settings.
 - [session_id_and_temporary_tables.ts](session_id_and_temporary_tables.ts) - creating a temporary table, which requires a session_id to be passed to the server.
 - [session_level_commands.ts](session_level_commands.ts) - using SET commands, memorized for the specific session_id.
+- [role.ts](role.ts) - using one more more roles, without explicit `USE` commands or session IDs
 
 ## How to run
 
--- a/examples/role.ts
+++ b/examples/role.ts
@@ -1,0 +1,23 @@
+import { createClient } from '@clickhouse/client' // or '@clickhouse/client-web'
+
+void (async () => {
+  const client = createClient({
+    role: 'role_name_1',
+  })
+
+  // with a role defined in the client configuration, all queries will use the specified role
+  await client.command({
+    query: `SELECT * FROM SECURED_TABLE`,
+  })
+
+  // one or more roles can be specified in a query as well, to override the role(s) set for the client
+  const rows1 = await client.query({
+    query: `SELECT * FROM VERY_SECURED_TABLE`,
+    format: 'JSONEachRow',
+    role: ['highly_privileged_role'],
+  })
+
+  console.log(await rows1.json())
+
+  await client.close()
+})()
--- a/packages/client-common/__tests__/integration/role.test.ts
+++ b/packages/client-common/__tests__/integration/role.test.ts
@@ -1,0 +1,359 @@
+import type { ClickHouseClient } from '@clickhouse/client-common'
+import { createTestClient, TestEnv, whenOnEnv } from '@test/utils'
+import { getTestDatabaseName, guid } from '../utils'
+import { createSimpleTable } from '../fixtures/simple_table'
+import { assertJsonValues, jsonValues } from '../fixtures/test_data'
+
+describe('role settings', () => {
+  let defaultClient: ClickHouseClient
+  let client: ClickHouseClient
+
+  let database: string
+  let username: string
+  let password: string
+  let roleName1: string
+  let roleName2: string
+
+  beforeAll(async () => {
+    defaultClient = createTestClient()
+    username = `clickhousejs__user_with_roles_${guid()}`
+    password = `CHJS_${guid()}`
+    roleName1 = `TEST_ROLE_${guid()}`
+    roleName2 = `TEST_ROLE_${guid()}`
+    database = getTestDatabaseName()
+
+    await defaultClient.command({
+      query: `CREATE USER ${username} IDENTIFIED WITH sha256_password BY '${password}' DEFAULT DATABASE ${database}`,
+    })
+    await defaultClient.command({
+      query: `CREATE ROLE IF NOT EXISTS ${roleName1}`,
+    })
+    await defaultClient.command({
+      query: `CREATE ROLE IF NOT EXISTS ${roleName2}`,
+    })
+    await defaultClient.command({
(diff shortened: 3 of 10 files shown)


examples/role.ts
```diff
@@ -0,0 +1,23 @@
...
+  const rows1 = await client.query({
+    query: `SELECT * FROM VERY_SECURED_TABLE`,
+    format: 'JSONEachRow',
+    role: ['highly_privileged_role'],
+  })
+
+  console.log(await rows1.json())
+
+  await client.close()
+})()
```
slvrtrn, line 124: We aim to have runnable examples; I suggest using the roles bootstrap from tests here.

It is also a good idea to:
- demonstrate expected failures with CH error codes (printing it in the console will suffice so that we can see that 497 ACCESS_DENIED, just like in some other examples, like `abort_request.ts`, `cancel_query.ts`, and probably some others)
- print the output rows like in all other examples (the result set should be consumed, as it holds an unreleased keep-alive socket, waiting for streaming to start; again, a misuse in an example - ResultSets should not be "dangling" in the code)

examples/role.ts
```diff
@@ -0,0 +1,23 @@
...
+
+void (async () => {
+  const client = createClient({
+    role: 'role_name_1',
+  })
+
+  // with a role defined in the client configuration, all queries will use the specified role
+  await client.command({
+    query: `SELECT * FROM SECURED_TABLE`,
+  })
```
slvrtrn, line 53: This should be a `query` call; now it is a misuse in an example, we want to avoid that

packages/client-common/__tests__/integration/role.test.ts
```diff
@@ -0,0 +1,359 @@
...
+
+    async function tryCreateTable(role?: string | Array<string>) {
+      const query = `
+        CREATE TABLE ${tableName}
+        (id UInt64, name String, sku Array(UInt8), timestamp DateTime)
+        ENGINE = MergeTree()
+        ORDER BY (id)
+      `
+      await client.command({ query, role })
+    }
```
slvrtrn, line 252: Nit: you could just use `createSimpleTable` like [here](https://github.com/ClickHouse/clickhouse-js/blob/bd262b9aea9ea509b945924d882e3715f4086cb6/packages/client-common/__tests__/integration/auth.test.ts#L51-L66)

examples/README.md
```diff
@@ -67,6 +67,7 @@ If something is missing, or you found a mistake in one of these examples, please
 - [default_format_setting.ts](default_format_setting.ts) - sending queries using `exec` method without a `FORMAT` clause; the default format will be set from the client settings.
 - [session_id_and_temporary_tables.ts](session_id_and_temporary_tables.ts) - creating a temporary table, which requires a session_id to be passed to the server.
 - [session_level_commands.ts](session_level_commands.ts) - using SET commands, memorized for the specific session_id.
+- [role.ts](role.ts) - using one more more roles, without explicit `USE` commands or session IDs
```
slvrtrn:
```suggestion
- [role.ts](role.ts) - using one or more roles without explicit `USE` commands or session IDs
```

examples/role.ts
```diff
@@ -0,0 +1,23 @@
+import { createClient } from '@clickhouse/client' // or '@clickhouse/client-web'
+
+void (async () => {
```
slvrtrn, line 8: Let's add a "See also" link to https://clickhouse.com/docs/en/interfaces/http#setting-role-with-query-parameters

packages/client-common/__tests__/unit/to_search_params.test.ts
```diff
@@ -78,6 +78,7 @@ describe('toSearchParams', () => {
         qaz: 'qux',
       },
       session_id: 'my-session-id',
+      role: ['my-role-1', 'my-role-2'],
```
slvrtrn, line 81: Let's add a test with a single role (as a string).

packages/client-common/src/utils/url.ts
```diff
@@ -78,5 +80,13 @@ export function toSearchParams({
...
   }
 
+  if (role && typeof role === 'string') {
+    params.set('role', role)
+  } else if (role && Array.isArray(role)) {
+    for (const r of role) {
+      params.append('role', r)
+    }
+  }
+
```
slvrtrn: Nit: maybe just a single undefined check will suffice?
```suggestion
  if (role) {
    if (typeof role === 'string') {
      params.set('role', role)
    } else if (Array.isArray(role)) {
      for (const r of role) {
        params.append('role', r)
      }
    }
  }

```

slvrtrn (changes requested):
Thanks for the contribution! Tests are good, too. There is one thing that we should fix before merging (the example); the rest are mostly nitpicks

packages/client-common/__tests__/integration/role.test.ts
pulpdrew: The problem I have with that approach is that it doesn't allow me to pass through a `role` setting to add to the `.command` call in `createTable()`. I could update `createSimpleTable` and `createTable` to accept optional `command` options if you'd like, or keep it as is, or if you had a different workaround in mind please let me know

slvrtrn: Ah yes, I see. It's fine to leave it like this then.

slvrtrn:
The CLA also needs to be signed before we merge it.

pulpdrew:
> The CLA also needs to be signed before we merge it.

Will do - I am working with my company to make sure I am not violating any of their open source policies by signing, but will sign as soon as I have approval.

examples/role.ts
pulpdrew: Updated the examples so that they run successfully, consume output rows, and print any error codes.

Example output:
Screenshot 2024-09-28 at 10 09 35 AM

examples/role.ts
pulpdrew: Updated all SELECT queries to use `.query` instead of `.command`

slvrtrn:
@pulpdrew, do you have any updates regarding the CLA? I'd like to merge this, but non-signed CLA is a blocker.

pulpdrew:
> @pulpdrew, do you have any updates regarding the CLA? I'd like to merge this, but non-signed CLA is a blocker.

Hey, apologies for the delay. My company is still reviewing the CLA. 🫠  They tell me they should have an update for me early next week. I will reply here with an ETA next week once I hear from them!

pulpdrew:
@slvrtrn I have signed the CLA, thanks for waiting!

slvrtrn (approved):

Issue #332, opened 2024-09-26
An error occurred in a middle of a select stream processing should be thrown


slvrtrn:
Min repro (note: `wait_end_of_query` setting is not used):

```ts
import { createClient } from '@clickhouse/client'

void (async () => {
  const client = createClient()
  let rowsCount = 0
  try {
    const resultSet = await client.query({
      query: `
        SELECT
          throwIf(number = 2, 'There was an error in the stream!') AS e,
          randomPrintableASCII(1) AS s
        FROM system.numbers
        LIMIT 2000
      `,
      format: 'JSONEachRow',
      clickhouse_settings: {
        max_block_size: '1',
      },
    })
    console.log(await resultSet.json())
  } catch (err) {
    console.error(`Got ${rowsCount} rows before an error:`, err)
  }
  await client.close()
})()
```

Prints:

```
[
  { e: 0, s: '}' },
  {
    exception: "Code: 395. DB::Exception: There was an error in the stream!: while executing 'FUNCTION throwIf(equals(__table1.number, 2_UInt8) :: 3, 'There was an error in the stream!'_String :: 2) -> throwIf(equals(__table1.number, 2_UInt8), 'There was an error in the stream!'_String) UInt8 : 0'. (FUNCTION_THROW_IF_VALUE_IS_NON_ZERO) (version 24.8.3.59 (official build))"
  }
]
```

However, the client should handle and throw during the stream consumption instead of providing the error as a row.

Caveat: the stream error representation depends on the format.

For example:
- `JSON`, `JSONCompact`, `JSONStrings`, `JSONCompactStrings` - the exception resides in the `exception` field and can be extracted from there
  ```
  {
	  "meta":
	  [
		  {
			  "name": "throwIf(equals(number, 2), 'There was an error in the stream!')",
			  "type": "UInt8"
		  },
		  {
			  "name": "randomPrintableASCII(2000)",
			  "type": "String"
		  }
	  ],
  
	  "data":
	  [
  
	  ],
  
	  "rows": 0,
  
[...]

PR #350, opened 2024-11-05
fix: check the X-ClickHouse-Exception-Code for err code

dermasmid:
## Summary
use the `X-ClickHouse-Exception-Code` header https://github.com/ClickHouse/ClickHouse/pull/8786

Related: #332 (similar issue can happen with inserts)

### DIFF
--- a/packages/client-common/src/utils/connection.ts
+++ b/packages/client-common/src/utils/connection.ts
@@ -35,7 +35,7 @@ export function withHttpSettings(
 
 export function isSuccessfulResponse(
   statusCode?: number,
-  headers: Record<string, string> = {},
+  headers: any = {},
 ): boolean {
   return Boolean(
     statusCode &&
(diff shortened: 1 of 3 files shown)


Moshe-Fromer (approved):

slvrtrn:
> fixes: https://github.com/ClickHouse/clickhouse-js/issues/332

It does not fix it for select streaming, as the headers will already be received; they are only for inserts.

dermasmid:
ok, so this solves a similar issue we had, should i open a separate issue for that?

slvrtrn:
> ok, so this solves a similar issue we had, should i open a separate issue for that?

No, this is fine. Thanks for the contribution!

slvrtrn (approved):

slvrtrn:
@dermasmid, is it possible to add an integration test for that? And one more question: in case of a CH error rejection, we _usually_ return an instance of `ClickHouseError`. Is it the same when `X-ClickHouse-Exception-Code` is present in the response?

dermasmid:
in theory the error should be parsed just fine, since it's using regex to find the error from the body: https://github.com/ClickHouse/clickhouse-js/blob/00af5c2cd90d7d37775158052e3e072be091c393/packages/client-common/src/error/parse_error.ts#L2

ill try to get a test in to

Issue #106, opened 2022-10-01
Invalid string length exception on large reads


genzgd:
Crashes when reading 50,000,000 numbers (at least on my Mac M1).

Code to reproduce (against localhost/local docker):

```javascript
const { createClient } = require('@clickhouse/client')

(async () => {
    const client = createClient();
    console.log(Date.now() / 1000)
    const result = await client.query({
        query: 'SELECT number FROM numbers(50000000)',
        format: "JSONCompactEachRow"
    });
    console.log(Date.now() / 1000);
    const stuff = await result.json();
    console.log(stuff.length);
    console.log(Date.now() / 1000);

})()
```

Exception:

```
1664586660.882
1664586660.895
AbortError: The operation was aborted
    at Object.destroyer (node:internal/streams/destroy:305:11)
    at createAsyncIterator (node:internal/streams/readable:1141:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  code: 'ABORT_ERR'
}
/Users/geoff/projects/nodetests/node_modules/@clickhouse/client/dist/utils/stream.js:16
        result += textDecoder.decode(chunk, { stream: true });
                              ^

RangeError: Invalid string length
    at getAsText (/Users/geoff/projects/nodetests/node_modules/@clickhouse/client/dist/utils/stream.js:16:31)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Rows.text (/Users/geoff/projects/nodetests/node_modules/@clickhouse/client/dist/rows.js:37:17)
    at async Rows.json (/Users/geoff/projects/nodetests/node_modules/@clickhouse/client/dist/rows.js:50:45)
    at async /Users/geoff/projects/nodetests/index.js:47:19

Node.js v18.9.0
```

slvrtrn:
Unfortunately, that seems to be a hard Node.js limitation. We cannot use strings larger than 512Mb at the moment. If I understand it correctly, it was increased from 256Mb to 1Gb in the past and then reduced to 512Mb.

See 
- https://stackoverflow.com/questions/44533966/v8-node-js-increase-max-allowed-string-length
- https://github.com/nodejs/help/issues/712#issuecomment-582709275
- https://github.com/v8/v8/commit/ea56bf5513d0cbd2a35a9035c5c2996272b8b728 

My tests showed that it indeed crashes at around 512Mb of size, with the resulting string's length around 536870888.

ck-clickhouse:
[...]

PR #357, opened 2024-11-27
Verify the response string length overflow and throw a specific error

slvrtrn:
## Summary

Related to https://github.com/ClickHouse/clickhouse-js/issues/106

According to [this commit](https://github.com/v8/v8/commit/ea56bf5513d0cbd2a35a9035c5c2996272b8b728), the client now throws a specific error if the result string length is more than the max allowed string length in V8, which is `2**29 - 24` on a 64-bit platform.

Previously, the exception was too ambiguous: 
- Node.js: `RangeError: Invalid string length`
- Browsers: `InternalError: allocation size overflow`

Now, the error message is supposed to be more clear.

## Checklist

### DIFF
--- a/packages/client-node/src/utils/stream.ts
+++ b/packages/client-node/src/utils/stream.ts
@@ -1,5 +1,8 @@
 import Stream from 'stream'
 
+// See https://github.com/v8/v8/commit/ea56bf5513d0cbd2a35a9035c5c2996272b8b728
+const MaxStringLength = Math.pow(2, 29) - 24
+
 export function isStream(obj: any): obj is Stream.Readable {
   return obj !== null && typeof obj.pipe === 'function'
 }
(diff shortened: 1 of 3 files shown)


packages/client-node/src/utils/stream.ts
```diff
@@ -1,5 +1,8 @@
 import Stream from 'stream'
 
+// See https://github.com/v8/v8/commit/ea56bf5513d0cbd2a35a9035c5c2996272b8b728
+const MaxStringLength = Math.pow(2, 29) - 24
```
mshustov, line 4: I'm just curious: is it relevant for all supported nodejs versions?

mshustov (approved):

packages/client-node/src/utils/stream.ts
slvrtrn: and also browsers, as it comes from the V8 itself

PR #359, opened 2024-11-28
Improvements to the Tuple and Map parameter binding

### LINKED ISSUE #358: Query fails to parse tuple when provided as an array
### Describe the bug

The query fails to work as expected when a query_param of type tuple is provided as an array (e.g., `['foo', 'bar']`). 

### Steps to reproduce
1. Use the code example to select the first tuple element:
2. Observe the error when running the query.

### Expected behaviour

The query should successfully return the first element of the tuple. 
On ClickHouse Cloud Console it's working with `('foo', 'bar')`

### Code example

```
const res = await client.query({
        query: `SELECT tupleElement({tupleParam: Tuple(String, String)}, 1) AS firstElement`,
        query_params: {
          tupleParam: ['foo', 'bar'],
        },
      })
```

### Error log

```
Error: Cannot parse input: expected '(' before: '[\'foo\',\'bar\']': value ['foo','bar'] cannot be parsed as Tuple(String, String) for query parameter 'tupleParam'. 
    at parseError (/Users/xx/xx/node_modules/packages/client-common/src/error/parse_error.ts:30:12)
    at ClientRequest.onResponse (/Users/xx/xx/node_modules/packages/client-node/src/connection/node_base_connection.ts:477:28)
    at processTicksAndRejections (node:internal/process/task_queues:95:5) 
 code: '27',
 type: 'CANNOT_PARSE_INPUT_ASSERTION_FAILED'
```

#### Environment
* Client version: 1.8.1
* Language version: node v20.11.1
* OS: macOS 14.3

#### ClickHouse server
* ClickHouse Server version: Cloud 24.8

slvrtrn:
## Summary

Closes #358 

- The client now exports a new wrapper class: `TupleParam`, as tuples need to be serialized as just strings like `('foo','bar')`; however, if it is provided like that in the current version, the single quotes will be escaped, which results in a query failure.

- Added a proper format for the JS `Map` type (it was missing previously), as specifying maps as just objects works only if the CH Map key is a String (cause the JS keys are always strings), and fails in case of Ints, etc.

## Checklist

### DIFF
--- a/packages/client-common/src/data_formatter/format_query_params.ts
+++ b/packages/client-common/src/data_formatter/format_query_params.ts
@@ -1,3 +1,7 @@
+export class TupleParam {
+  constructor(public readonly values: any[]) {}
+}
+
 export function formatQueryParams(
   value: any,
   wrapStringInQuotes = false,
@@ -36,8 +40,7 @@ export function formatQueryParams(
   }
 
   if (Array.isArray(value)) {
-    const formatted = value.map((v) => formatQueryParams(v, true))
-    return `[${formatted.join(',')}]`
+    return `[${value.map((v) => formatQueryParams(v, true)).join(',')}]`
   }
 
   if (value instanceof Date) {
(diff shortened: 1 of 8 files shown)


packages/client-common/src/data_formatter/format_query_params.ts
```diff
@@ -51,6 +54,21 @@ export function formatQueryParams(
       : `${unixTimestamp}.${milliseconds.toString().padStart(3, '0')}`
   }
 
+  if (value instanceof TupleParam) {
```
mshustov, line 57: `Array` seems to have the same format. Can we accept JS `Array` as a valid data structure for `Tuple`? It would improve DX since no special class `TupleParam` is needed

slvrtrn: Array is `[A,B,...]`, not `(A,B,...)`

```
curl "http://localhost:8123?param_val=('foo','bar')" --data-binary "SELECT {val:Array(String)} JSONEachRow"
Code: 130. DB::Exception: Array does not start with '[' character: value ('foo','bar') cannot be parsed as Array(String) for query parameter 'val'. (CANNOT_READ_ARRAY_FROM_TEXT) (version 24.8.3.59 (official build))
```

```
curl "http://localhost:8123?param_val=\['foo','bar'\]" --data-binary "SELECT {val:Array(String)} FORMATJSONEachRow"
['foo','bar']
```

mshustov (approved):

PR #360, opened 2024-11-29
JWT auth support

### LINKED ISSUE #270: Add JWT support
After https://github.com/ClickHouse/ClickHouse/issues/17270

slvrtrn:
## Summary

- Support for JWT tokens auth is added (an upcoming feature in ClickHouse Cloud). Closes #270 
- Top-level `username` and `password` options in the config are now deprecated; the `auth` object should be used instead.
- Supported via the URL configuration as the `access_token` param.
- It is possible to override JWT token for a particular request, too.
- The related tests are ignored until we have an instance with JWT support.

The URL config example needs to be updated - maybe later as a follow-up.

## Checklist


packages/client-common/src/clickhouse_types.ts
```diff
@@ -46,6 +46,25 @@ export interface ProgressRow {
...
+export interface CredentialsAuth {
+  username?: string
+  password?: string
+}
+
+export interface JWTAuth {
+  access_token: string
+}
+
+export type Auth = CredentialsAuth | JWTAuth
```
slvrtrn: Shall we name it `ClickHouseAuth` instead? 

I will also add a doc entry.

packages/client-common/src/config.ts
```diff
@@ -397,6 +438,9 @@ export function loadConfigOptionsFromURL(
             }
             config.keep_alive.enabled = booleanConfigURLValue({ key, value })
             break
+          case 'access_token':
+            config.auth = { access_token: value }
```
slvrtrn, line 445: Is it OK to leave it as just `access_token` (similarly to what we currently have with `username` and `password`, so it is somewhat consistent), or make a key that resembles the full path (`auth_access_token`)?

slvrtrn: The URL config example needs to be updated.

slvrtrn:
> 82.9% Coverage on New Code

It should increase once the related tests are enabled.

packages/client-common/src/config.ts
mshustov: what is the benefit of the full path `auth_access_token`?

packages/client-node/src/connection/node_base_connection.ts
```diff
@@ -258,19 +265,28 @@ export abstract class NodeBaseConnection
...
-      return {
-        ...this.defaultHeaders,
+    }
+    if (this.params.set_basic_auth_header) {
+      if (isCredentialsAuth(params?.auth)) {
+        return {
+          ...this.defaultHeaders,
+          Authorization: `Basic ${Buffer.from(`${params.auth.username}:${params.auth.password}`).toString('base64')}`,
+        }
+      } else {
```
mshustov, line 280: optional: can we move `Auth header` calculation logic to a separate function and test separately? It's hard to reason about the flow with a few levels of `if`
```js
if (isJWTAuth(params?.auth)) {...}
if (this.params.set_basic_auth_header) {
      if (isCredentialsAuth(params?.auth)) { ... } 
      else {...}
}
```

packages/client-node/src/connection/node_https_connection.ts
```diff
@@ -24,10 +27,24 @@ export class NodeHttpsConnection extends NodeBaseConnection {
...
-        ...this.defaultHeaders,
-        'X-ClickHouse-User': params?.auth?.username ?? this.params.username,
-        'X-ClickHouse-Key': params?.auth?.password ?? this.params.password,
+      if (this.params.auth.type === 'JWT') {
+        throw new Error(
+          'JWT auth is not supported with HTTPS connection using custom certificates',
+        )
+      }
+      let headers: Http.OutgoingHttpHeaders
+      if (isCredentialsAuth(params?.auth)) {
```
mshustov, line 36: It's not the same behavior. With the current change, both params must be provided. I'm okay with it. Let's just add it to the changelog

mshustov (approved):

packages/client-node/src/connection/node_https_connection.ts
slvrtrn: Previously, it was also required to provide both username and password to override the credentials. The behavior should be exactly the same.

packages/client-common/src/config.ts
slvrtrn: A bit more consistent with others like `keep_alive_enabled` and so on. But since this is a "core" authentication-related setting, and the name is also not ambiguous, it is probably better to keep it without this prefix for simplicity.

Issue #370, opened 2025-01-01
support for session_timeout parameter


otan:
### Use case

Using temporary tables, our jobs (unfortunately) can have a long delay sometimes that requires a session to be idle for a while.

The docs make mention of `default_session_timeout` and `session_timeout`: https://clickhouse.com/docs/en/interfaces/http. 

> Similarly, you can use ClickHouse sessions in the HTTP protocol. To do this, you need to add the session_id GET parameter to the request. You can use any string as the session ID. By default, the session is terminated after 60 seconds of inactivity. To change this timeout, modify the default_session_timeout setting in the server configuration, or add the session_timeout GET parameter to the request. To check the session status, use the session_check=1 parameter. Only one query at a time can be executed within a single session.

However, I cannot find `session_timeout` on the client.
I also cannot find where to modify `default_session_timeout` on the server config for ClickHouse cloud as it does not exist in `system.settings`.

### Describe the solution you'd like

Exposing session_timeout in the client is most ideal.

### Describe the alternatives you've considered

Not sure any besides writing our own http client.

slvrtrn:
I think it should work if you just put it to `clickhouse_settings`, e.g.

```ts
const rs = client.query({
  query: 'SELECT ...',
  clickhouse_settings: {
    session_timeout: '3600',
    session_check: '1',
  }
})
```

slvrtrn:
To simplify TS usage, makes sense to add it to https://github.com/ClickHouse/clickhouse-js/blob/main/packages/client-common/src/settings.ts#L1584-L1587

slvrtrn:
> would need to disable lint if it exist in typescript, but will give it a shot. 

TS compiler shouldn't complain, see `Record` here: https://github.com/ClickHouse/clickhouse-js/blob/9594c1843c1a941a4c83e4ec621505b165752890/packages/client-common/src/settings.ts#L1589-L1591

> is the unit seconds or milliseconds?

it's in seconds: https://github.com/ClickHouse/ClickHouse/blob/fe35e6d3def4c73755e7bd20b65ebe2a2e0a52a4/src/Server/HTTPHandler.cpp#L126-L148

See also: https://github.com/ClickHouse/ClickHouse/blob/fe35e6d3def4c73755e7bd20b65ebe2a2e0a52a4/programs/server/config.xml#L997-L998

the docs should be updated, it's not 100% clear from there.

PR #372, opened 2025-01-03
Add `session_timeout` and `session_check` settings typings

slvrtrn:
## Summary

Related to #370. It is possible to use these settings anyway cause the typings are not very strict; however, it is more convenient to have it there for autocomplete. 

Added some docs entries there as well.

See also https://clickhouse.com/docs/en/interfaces/http#default-database

### DIFF
--- a/packages/client-common/src/settings.ts
+++ b/packages/client-common/src/settings.ts
@@ -1596,8 +1596,11 @@ interface ClickHouseHTTPSettings {
   /** By default, the session is terminated after 60 seconds of inactivity
    *  This is regulated by the `default_session_timeout` server setting. */
   session_timeout: UInt64
-  /** To check the session status, you can use this setting.
-   *  If a session is expired or cannot be found, server returns `SESSION_NOT_FOUND` with error code 372. */
+  /** You can use this setting to check the session status before executing the query.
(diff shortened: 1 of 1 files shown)


packages/client-common/src/settings.ts
```diff
@@ -1581,9 +1581,24 @@ interface ClickHouseServerSettings {
...
   wait_end_of_query: Bool
+  /** Format to use if a SELECT query is executed without a FORMAT clause.
+   *  Only useful for the {@link ClickHouseClient.exec} method,
+   *  as {@link ClickHouseClient.query} method always attaches this clause. */
   default_format: DataFormat
+  /** By default, the session is terminated after 60 seconds of inactivity
+   *  This is regulated by the `default_session_timeout` server setting. */
+  session_timeout: UInt64
+  /** To check the session status, you can use this setting.
+   *  If a session is expired or cannot be found, server returns `SESSION_NOT_FOUND` with error code 372. */
```
mshustov: Should we add a warning that the session doesn't work well in multi-node env?

mshustov (approved):

packages/client-common/src/settings.ts
slvrtrn: Added.

Issue #394, opened 2025-02-24
Ability to supply Open Telemetry traceparent with request


DylanRJohnston:
### Use case
Being able to associate the traces in [system.opentelemetry_span_log](https://clickhouse.com/docs/operations/system-tables/opentelemetry_span_log) with a client query. As documented [here](https://clickhouse.com/docs/operations/opentelemetry)

### Describe the solution you'd like

Clickhouse documentation states that 

> ClickHouse accepts trace context HTTP headers, as described by the [W3C recommendation](https://www.w3.org/TR/trace-context/). It also accepts trace context over a native protocol that is used for communication between ClickHouse servers or between the client and server.

However there doesn't appear to be any way to set these given the current client interface. You can set the request headers once during `createClient` but for the headers to be useful they need to be settable for each query, as each query will have a different traceparent.

```ts
const result = await client.query({ query, query_params, headers: { traceparent }})
```

DylanRJohnston:
I've temporarily found a work around for this using some gross Proxy magic and intercepting the serialisation of the header with `Symbol.toPrimitive` and then setting the module global `currentSpan` just before calling `client.query`. This is very fragile though as it relies upon the header being serialised without yielding back to the event loop potentially causing `currentSpan` to be set to a different value.

```ts
 const client = createClient({
      url: process.env?.["CLICKHOUSE_URL"],
      username: process.env?.["CLICKHOUSE_USERNAME"],
      password: process.env?.["CLICKHOUSE_PASSWORD"],
      http_headers: {
        traceparent: new Proxy({}, {
          get: (target, prop) => {
            if (prop !== Symbol.toPrimitive) {
              return Reflect.get(target, prop);
            }

            return () =>
              `00-${currentSpan!.traceId}-${currentSpan!.spanId}-0${Number(
                currentSpan!.sampled ? TraceFlags.SAMPLED : TraceFlags.NONE
              ).toString(16)}`;
          },
        }) as any,
      },
```

DylanRJohnston:
Following up from the above hack with `Symbol.toPromitive`, after further testing it is not at all reliable, I guess we do yield back to the event loop in between setting the global `currentSpan` and the header being serialised. I've opened a PR #395, that adds the ability to specify the headers when querying.

PR #395, opened 2025-02-26
feat(node): add paramters to forward OTEL Headers to Clickhouse

DylanRJohnston:
## Summary
Addresses #394, Clickhouse supports reading the OTEL HTTP headers from the connection

## Checklist
- [x] Unit tests covering the common scenarios were added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 1.10.2
+
+## Improvements
+- Added `opentelemetry_headers` option to the client configuration. ([#394](https://github.com/ClickHouse/clickhouse-js/issues/374))
+
 # 1.10.1 (Common, Node.js, Web)
 
 ## Bug fixes
--- a/packages/client-common/src/client.ts
+++ b/packages/client-common/src/client.ts
@@ -43,6 +43,12 @@ export interface BaseQueryParams {
         password: string
       }
     | { access_token: string }
+  /** OpenTelemetry headers to propagate to ClickHouse.
+   *  @default undefined (no override) */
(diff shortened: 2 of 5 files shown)


DylanRJohnston:
@slvrtrn that makes sense, but what about the existing auth configuration parameters that get turned into headers? Should custom headers override them, or visa versa?

DylanRJohnston:
Also I had a look at the failing integration tests it seems that the CI environment is misconfigured?

mshustov:
> I think it makes more sense to just allow overriding headers for a particular request in all methods. It also looks like that's what initially was proposed in the code snippet in the issue.

👍 

> what about the existing auth configuration parameters that get turned into headers? Should custom headers override them, or visa versa?

I'd vote for allowing users to override any headers (with a debug log record emitted that an override has happened) - it gives more power and flexibility to the end user. For example, one can implement multi-tenant runtime and mange users on the app level

slvrtrn:
> Should custom headers override them, or visa versa?

Let's make the same logic to the default headers in the connection class: https://github.com/ClickHouse/clickhouse-js/blob/main/packages/client-node/src/connection/node_base_connection.ts#L97-L101 (+ `buildRequestHeaders`)

e.g. can override all except `Connection`, `User-Agent` and `Authorization`. For the authorization header, there is already a separate option in the query settings.

> it seems that the CI environment is misconfigured?

Yep, it cannot run the tests with a Cloud instance if a PR is created by an external contributor.

slvrtrn:
Closing this, as it is now included in https://github.com/ClickHouse/clickhouse-js/pull/402

PR #402, opened 2025-03-11
Allow to override HTTP headers on the method level

### LINKED ISSUE #394: Ability to supply Open Telemetry traceparent with request

slvrtrn:
## Summary

Based on #395; allows to override the HTTP headers on the method level for both Node.js and Web implementations. 

Closes #394, as it will be possible to specify the headers as follows:

```ts
const resultSet = await client.query({
  http_headers: {
    traceparent: '...',
    tracestate: '...',
  },
)
```

## Checklist

### DIFF
--- a/packages/client-web/__tests__/integration/web_client.test.ts
+++ b/packages/client-web/__tests__/integration/web_client.test.ts
@@ -1,5 +1,5 @@
+import { getHeadersTestParams } from '@test/utils/parametrized'
 import { createClient } from '../../src'
-import type { WebClickHouseClient } from '../../src/client'
 
 describe('[Web] Client', () => {
   let fetchSpy: jasmine.Spy<typeof window.fetch>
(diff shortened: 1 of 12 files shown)


packages/client-web/__tests__/integration/web_client.test.ts
```diff
@@ -16,46 +16,85 @@ describe('[Web] Client', () => {
...
+            FromMethod: 'bar',
+          })
+
+        await param.methodCall({ FromInstance: 'bar' })
+        expect(getFetchRequestInit(fetchCalls++).headers)
+          .withContext(
+            `${param.methodName}: overrides HTTP headers from the instance with the values from the method call`,
+          )
+          .toEqual({
+            ...defaultHeaders,
```
mshustov, line 61: can we add a test that Authz header is not overriden?

packages/client-web/__tests__/integration/web_client.test.ts
```diff
@@ -16,46 +16,85 @@ describe('[Web] Client', () => {
...
+      let fetchCalls = 1
+      const testParams = getHeadersTestParams(client)
+      for (const param of testParams) {
+        await param.methodCall({ FromMethod: 'bar' })
+        expect(getFetchRequestInit(fetchCalls++).headers)
+          .withContext(
+            `${param.methodName}: merges custom HTTP headers from both method and instance`,
+          )
+          .toEqual({
+            ...defaultHeaders,
```
mshustov, line 52: can we add a test that the Authorization header is not overridden when provided in `http_headers`? The same for nodejs implementation

mshustov (approved):

PR #403, opened 2025-03-20
Support 32-bit architecture in stream

shevchenkonik:
## Summary
The PR introduces two improvements related to the `stream` utility:

1. Replaces the hardcoded value with the `MAX_STRING_LENGTH` constant from [node.buffer](https://nodejs.org/api/buffer.html#bufferconstantsmax_string_length). The issue with the hardcoded value is that it doesn't account for 32-bit architectures and the function couldn't detect the limit.
2. Replaces the use of `any` with `unknown` in the `isStream` function.

I didn't add tests since the main flow remains unchanged, and the update addresses 32-bit architectures only.

### DIFF
--- a/packages/client-node/src/utils/stream.ts
+++ b/packages/client-node/src/utils/stream.ts
@@ -3,8 +3,15 @@ import { constants } from 'buffer'
 
 const { MAX_STRING_LENGTH } = constants
 
-export function isStream(obj: any): obj is Stream.Readable {
-  return obj !== null && typeof obj.pipe === 'function'
+export function isStream(obj: unknown): obj is Stream.Readable {
+  return (
+    typeof obj === 'object' &&
+    obj !== null &&
+    'pipe' in obj &&
+    typeof obj.pipe === 'function' &&
+    'on' in obj &&
+    typeof obj.on === 'function'
+  )
 }
 
 export async function getAsText(stream: Stream.Readable): Promise<string> {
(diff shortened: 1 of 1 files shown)


slvrtrn:
Thanks for the contribution. 

I don't think we can simply use `constants` because previously, when the "max string length" value was unbound, it was throwing a cryptic error that was difficult for the end user to understand, and that happened even on a 64-bit system: see https://github.com/ClickHouse/clickhouse-js/pull/357; that's why it is limited to exactly that value. 

Perhaps we can detect that the client runs on a 32-bit system and use another value, which is probably also lower than the value provided in the buffer constants.

shevchenkonik:
Hello @slvrtrn, thank you for the answer.

The [introduced variable](https://github.com/ClickHouse/clickhouse-js/blob/4affe3432b99c5189ec8fa0cd9f7d6923d812d66/packages/client-node/src/utils/stream.ts#L4) into `clickhouse-js` equals to `536870888` and it works perfectly with 64-bit architectures. Unfortunately, this value is higher than the 32-bit architecture supports in terms of the v8 engine. V8 engine value for 32-bit is [268435440](https://github.com/v8/v8/blob/3a3130dfd68ed2945d94328bbefea626190d1f1d/include/v8-primitive.h#L127C46-L127C60). It means that we can't catch issue by [this condition](https://github.com/ClickHouse/clickhouse-js/blob/main/packages/client-node/src/utils/stream.ts#L16) when the codebase is running into 32-bit. 

I have checked node versions from 16 (because package.json says this library supports only ^16), and all major node versions have the `MAX_STRING_LENGTH` variable equal to `536870888`.

The changes in this PR haven't changed the flow of the function, and it should catch this error as before. However, using variables from `node` helps us catch the issue for both use cases, not only for 64-bit.

PR #422, opened 2025-05-14
Added setting lightweight_deletes_sync

pratimapatel2008:
## Summary
Added setting  `lightweight_deletes_sync` so that it can be used with JS library.

### DIFF
--- a/packages/client-common/src/settings.ts
+++ b/packages/client-common/src/settings.ts
@@ -761,6 +761,8 @@ interface ClickHouseServerSettings {
   legacy_column_name_of_tuple_literal?: Bool
   /** Limit on read rows from the most 'end' result for select query, default 0 means no limit length */
   limit?: UInt64
+  /** Controls the synchronicity of lightweight DELETE operations. It determines whether a DELETE statement will wait for the operation to complete before returning to the client. */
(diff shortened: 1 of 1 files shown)


packages/client-common/src/settings.ts
```diff
@@ -1681,6 +1681,8 @@ export interface MergeTreeSettings {
   index_granularity_bytes?: UInt64
   /** Retry period for table initialization, in seconds. */
   initialization_retry_period?: Seconds
+  /** Controls the synchronicity of lightweight DELETE operations. It determines whether a DELETE statement will wait for the operation to complete before returning to the client. */
+  lightweight_deletes_sync?: UInt64
```
slvrtrn: This is not a `MergeTreeSetting`; it should go to `ClickHouseServerSettings` so it's available as a typing in the `clickhouse_settings` argument to the instance or the methods.

slvrtrn (changes requested):

packages/client-common/src/settings.ts
pratimapatel2008: Thank you, I have moved the settings to `ClickHouseServerSettings`

PR #442, opened 2025-07-15
An option to use a simple SELECT query for ping checks instead of `/ping` endpoint

### LINKED ISSUE #440: Use a `SELECT` query instead of `/ping` endpoint in the `Client.ping` method
It is more convenient and uniform to just use something like

```sql
SELECT 'ping'
```

instead of sending an HTTP request to the `/ping` endpoint, as many users also want to validate authentication with this request.

Cons `/ping` endpoint:
- does not validate credentials
- does not implement CORS

slvrtrn:
Closes #440 

`ping` now always uses `SELECT 'ping'` instead of the endpoint for uniform behavior between Node.js and Web versions, as well as it can be used to check if credentials are valid now.


packages/client-node/src/connection/node_base_connection.ts
```diff
@@ -108,15 +108,21 @@ export abstract class NodeBaseConnection
...
+        database: undefined,
+        query: PingQuery,
+        query_id,
+      })
       const { stream } = await this.request(
         {
           method: 'GET',
-          url: transformUrl({ url: this.params.url, pathname: '/ping' }),
+          url: transformUrl({ url: this.params.url, searchParams }),
+          query: PingQuery,
```
mshustov, line 119: So, it's a breaking change since now ping validates the credentials? Can we add it via a flag as you proposed earlier? We can remove the flag in a major release. let's start tracking these breaking changes somewhere

slvrtrn: @mshustov, should be fixed

PR #457, opened 2025-09-03
feat(Infra): Upgrade eslint to v9

### LINKED ISSUE #451: Upgrade ESLint to 9.x
Automatic migration from 8.x to 9.x they suggest does not work as expected. While it executes properly:

```
npx @eslint/migrate-config .eslintrc.json 
Need to install the following packages:
@eslint/migrate-config@1.5.3
Ok to proceed? (y) y


Migrating .eslintrc.json
Also importing your .eslintignore file

Wrote new config to ./eslint.config.mjs

You will need to install the following packages to use the new config:
- globals
- @eslint/js
- @eslint/eslintrc

You can install them using the following command:

npm install globals @eslint/js @eslint/eslintrc -D

The following messages were generated during migration:
- The 'node' environment is used, but the sourceType is 'module'. Using sourceType 'module'. If you want to use CommonJS modules, set the sourceType to 'commonjs'.
```

Not everything is moved from JSON correctly:

```
✗ npm run lint

> clickhouse-js@0.0.0 lint
> eslint . --ext .ts

(node:551532) ESLintIgnoreWarning: The ".eslintignore" file is no longer supported. Switch to using the "ignores" property in "eslint.config.js": https://eslint.org/docs/latest/use/configure/migration-guide#ignoring-files
(Use `node --trace-warnings ...` to show where the warning was created)

Oops! Something went wrong! :(

ESLint: 9.33.0

file:///home/serge/work/clickhouse-js/eslint.config.mjs?mtime=1755171287375:4
import expectType from 'eslint-plugin-expect-type'
       ^^^^^^^^^^
SyntaxError: The requested module 'eslint-plugin-expect-type' does not provide an export named 'default'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:228:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:335:5)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:647:26)
    at async dynamicImportConfig (/home/serge/work/clickhouse-js/node_modules/eslint/lib/config/config-loader.js:186:17)
    at async loadConfigFile (/home/serge/work/clickhouse-js/node_modules/eslint/lib/config/config-loader.js:276:9)
    at async ConfigLoader.calculateConfigArray (/home/serge/work/clickhouse-js/node_modules/eslint/lib/config/config-loader.js:589:23)
    at async #calculateConfigArray (/home/serge/work/clickhouse-js/node_modules/eslint/lib/config/config-loader.js:743:23)
    at async directoryFilter (/home/serge/work/clickhouse-js/node_modules/eslint/lib/eslint/eslint-helpers.js:276:5)
    at async NodeHfs.<anonymous> (file:///home/serge/work/clickhouse-js/node_modules/@humanfs/core/src/hfs.js:586:29)
    at async NodeHfs.walk (file:///home/serge/work/clickhouse-js/node_modules/@humanfs/core/src/hfs.js:614:3)
```

eryue0220:
## Summary
Upgrade the eslint to v9 to fixed #451.


packages/client-common/__tests__/integration/abort_request.test.ts
```diff
@@ -130,7 +130,7 @@ describe('abort request', () => {
 
 async function assertActiveQueries(
   client: ClickHouseClient,
-  assertQueries: (queries: Array<{ query: string }>) => boolean,
+  assertQueries: (queries: { query: string }[]) => boolean,
```
slvrtrn: Arguably, Array is more readable there, and in the method definitions ReadonlyArray has some semantic value. Can we disable this rule/rewrite somehow?

tsconfig.json
```diff
@@ -22,9 +22,18 @@
...
-  "include": ["./packages/**/src/**/*.ts"]
+  "include": [
+    "./packages/**/src/**/*.ts",
+    "./packages/**/__tests__/**/*.ts",
+    ".build/**/*.ts",
+    ".scripts/**/*.ts",
+    "examples/**/*.ts",
+    "benchmarks/**/*.ts",
+    "__tests__/**/*.ts"
+  ]
```
slvrtrn: Can we leave the main `tsconfig.json` as it was? We have `tsconfig.all.json` already. 
Perhaps we can merge `tsconfig.dev.json` and `tsconfig.all.json` together instead.

packages/client-common/__tests__/integration/abort_request.test.ts
eryue0220: sure, or we can use `Array<{}>`?

tsconfig.json
eryue0220: updated.

packages/client-common/__tests__/integration/abort_request.test.ts
slvrtrn: At the second glance, for this particular case, it does not really matter, cause this is a test. However, I'd say just disabling the rewrite or warning for using Array/ReadonlyArray is the preferred approach; I don't see anything bad in using these types, and as I mentioned, sometimes they reflect the intention better, especially in the public interfaces.

slvrtrn: So let's remove the "array" rule, and revert the changes from `Array<T>` -> `T[]`
Then we can merge!

eryue0220: Now all `T[]` in the codebase are processed as `Array<T>`, I think it may be better to unify the style.

slvrtrn: > Now all T[] in the codebase are processed as Array<T>, I think it may be better to unify the style.

I'd prefer to just remove the rule and leave it as it was before this PR. This fix seems to be unrelated; can be addressed in a separate PR, but I don't see the need tbh.

eryue0220: Got it. I'll remove these change first and separate it into another PR.

eryue0220: Updated.

PR #462, opened 2025-09-19
feat(example): add example about advanced streaming insert.

### LINKED ISSUE #390: Advanced example with streaming INSERT
We currently provide examples for streaming from file https://github.com/ClickHouse/clickhouse-js/blob/72f490a6faf894deb723f2c6d5b7cdee75177def/examples/node/insert_file_stream_ndjson.ts

However, the examples don't cover a use case when a user app pushes to the `values` stream and should handle the backpressure mechanism correctly. If user app doesn't implement backpressure handling, it might lead to problems like these  https://github.com/ClickHouse/clickhouse-js/issues/135

eryue0220:
## Summary
This pr is try to fix #390 

## Note
This pr only add example codes, no any other feature or bug fix changes.

## Checklist


examples/node/insert_streaming_backpressure_simple.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  try {
+    console.log('Starting streaming insert with backpressure demonstration...')
+
+    const insertPromise = client.insert({
+      table: tableName,
+      values: dataStream,
+      format: 'JSONEachRow',
+      clickhouse_settings: {
+        // Use async inserts to handle streaming data more efficiently
```
slvrtrn, line 111: ~Does async insert (especially with wait) make a difference if we are streaming the records?~
EDIT: yes, it actually does.

examples/node/insert_streaming_with_backpressure.ts
```diff
@@ -0,0 +1,300 @@
...
+
+  console.log('Setting up table...')
+
+  await client.command({
+    query: `DROP TABLE IF EXISTS ${tableName}`,
+  })
+
+  await client.command({
+    query: `
+      CREATE TABLE ${tableName}
```
slvrtrn: Nit: just `CREATE OR REPLACE` should be sufficient

eryue0220: Updated.

examples/node/insert_streaming_with_backpressure.ts
```diff
@@ -0,0 +1,300 @@
...
+  // Called when the stream is ready to accept more data (backpressure resolved)
+  _read() {
+    if (this.#streamPaused && this.#pendingData.length > 0) {
+      // Process buffered data when backpressure is resolved
+      this.#streamPaused = false
+
+      // Push all pending data, but stop if we hit backpressure again
+      while (this.#pendingData.length > 0 && !this.#streamPaused) {
+        const data = this.#pendingData.shift()
+        if (!data || !this.#pushData(data)) {
```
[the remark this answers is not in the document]
eryue0220: It's not right, because the `data` might be `undefined`.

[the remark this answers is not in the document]
peter-leonov-ch: An interesting one. Don't we want the `undefined` or `""` value to also be forwarded through the pipe?

peter-leonov-ch (approved):
Thank you for your contribution! Streaming in Node.js is relatively new and a good example is always of value 👍

PR #470, opened 2025-10-15
Add support for zstd responses

ArctypeZach:


slvrtrn:
Let's revisit it as part of https://github.com/ClickHouse/clickhouse-js/issues/120 later in Q4 or Q1.

PR #542, opened 2026-02-09
Always resolve while draining the stream

peter-leonov-ch:
## Summary

Make sure that an already ended stream does not hang waiting for the `end` event.

While writing tests for `drainStream()` it turned out that `for await` covers all the tests cases.

## Checklist


packages/client-node/src/connection/stream.ts
```diff
@@ -4,35 +4,8 @@ import type Stream from 'stream'
...
-
-    stream.on('data', dropData)
-    stream.on('end', onEnd)
-    stream.on('error', onError)
-    stream.on('close', onClose)
-  })
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  for await (const _ of stream) {
+    // consume the stream
+  }
```
peter-leonov-ch, line 24: @slvrtrn this is too good to be true 😅 Is there anything I'm missing here?

mshustov: can it throw now? there used to be `error` handler as well

peter-leonov-ch: Yes it can. I've only dared to touch this piece after adding a bunch of unit tests for `drainStream()`. On the way I've fixed a hanging state when `drainStream()` receives a closed stream which does not emit the `end` event. This case is covered by the iterator out of the box.

The very reason I've tried the iterator here is that the pattern "check the state => add listeners => listen => remove listeners => resolve / reject" reminded about the even stream => async iterator wrappers in general. 

Stateful iterators handle both the streaming bit that we need here and the burdensome init / teardown stages.

peter-leonov-ch: Now if we wanted to silence the errors we'd just wrap the loop in a try/catch. It's also easier to measure time inside the function by wrapping the loop in a "span" or `console.time()`-like construct.

mshustov (approved):
LGTM. I'd wait for

packages/client-node/src/connection/stream.ts
```diff
@@ -18,62 +18,10 @@ export interface Context {
...
-    }
-
-    stream.on('data', dropData)
-    stream.on('end', onEnd)
-    stream.on('error', onError)
-    stream.on('close', onClose)
-  })
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  for await (const _ of stream) {
+    // consume the stream
```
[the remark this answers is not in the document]
peter-leonov-ch: Re-requesting review.

peter-leonov-ch:
Tests are still failing. Let's address this for the new `fetch()` centric implementation where we'd be forced to use modern language features.

PR #584, opened 2026-02-27
Add use_multipart_params option for large query parameters

jzila:
## Summary

- Adds a `use_multipart_params` client config option (default: `false`) that sends query parameters as `multipart/form-data` parts instead of URL query string entries
- When enabled, `query()` moves the SQL into a `query` multipart part and each `param_*` entry into its own named part, avoiding URL length limits that cause HTTP 400 on ClickHouse Cloud with large arrays (25K+ values)
- Uses the multipart format supported by ClickHouse since https://github.com/ClickHouse/ClickHouse/pull/40055
- Builds multipart bodies natively with `Buffer.concat()` — no new dependencies
- All non-param URL search params (database, query_id, settings, session_id, role) remain in the URL as-is

## Test plan

- [x] 7 unit tests for `buildMultipartBody()`: query part, param naming, array formatting, boundary structure, ordering, Buffer type, special character escaping
- [x] 7 unit tests for connection-level integration: params removed from URL, Content-Type header set with boundary, non-param search params preserved in URL, fallback to normal behavior when params are empty/undefined, default (non-multipart) behavior unchanged
- [x] All 77 existing Node.js unit tests pass
- [x] All 155 existing common unit tests pass
- [ ] Manual E2E test with ClickHouse instance using large array params


packages/client-common/src/utils/multipart.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (!SAFE_PARAM_NAME.test(key)) {
+      throw new Error(
+        `Invalid query parameter name: "${key}". ` +
+          'Parameter names must match /^[A-Za-z0-9_]+$/.',
+      )
+    }
+    const formattedValue = formatQueryParams({ value })
+    parts.push(
+      `--${boundary}\r\n` +
+        `Content-Disposition: form-data; name="param_${key}"\r\n` +
```
peter-leonov-ch: IMO, we better not hardcode the ClickHouse query domain into this function. Instead let's just implement a `Record<string, string> => string` function that could be used like this:

```ts
buildMultipartBody({
  query,
  ...query_params
}, boundary)
```

packages/client-common/src/utils/multipart.ts
```diff
@@ -0,0 +1,61 @@
...
+}): string {
+  const parts: string[] = []
+
+  // Query part
+  parts.push(
+    `--${boundary}\r\n` +
+      `Content-Disposition: form-data; name="query"\r\n` +
+      `\r\n` +
+      `${query}\r\n`,
+  )
```
peter-leonov-ch: Not sure I can see a reason for special casing the `query` itself, but I might be missing something.

packages/client-common/src/utils/multipart.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  // Parameter parts
+  for (const [key, value] of Object.entries(query_params)) {
+    if (!SAFE_PARAM_NAME.test(key)) {
+      throw new Error(
+        `Invalid query parameter name: "${key}". ` +
+          'Parameter names must match /^[A-Za-z0-9_]+$/.',
```
peter-leonov-ch: It should be handy to just stringify the regexp itself:

```suggestion
          `Parameter names must match ${SAFE_PARAM_NAME}.`,
```

packages/client-common/__tests__/unit/config.test.ts
```diff
@@ -375,6 +375,7 @@ describe('config', () => {
           parse: JSON.parse,
           stringify: JSON.stringify,
         },
+        use_multipart_params: false,
```
peter-leonov-ch: Seems like this and the following explicit parameter passing are extra now, consider removing.

packages/client-common/__tests__/unit/multipart.test.ts
```diff
@@ -0,0 +1,148 @@
...
+        `SELECT * FROM t WHERE x = {val:String}\r\n` +
+        `--${boundary}\r\n` +
+        `Content-Disposition: form-data; name="param_val"\r\n` +
+        `\r\n` +
+        `hello\r\n` +
+        `--${boundary}--\r\n`,
+    )
+  })
+
+  it('should build a multipart body with multiple params', () => {
```
peter-leonov-ch: I'd suggest to stick to the snapshot-like testing like the test above. Even better if we could use the native support for inline snapshots provided by Vitest.

packages/client-common/src/config.ts
```diff
@@ -261,6 +268,7 @@ export function getConnectionParams(
       ...defaultJSONHandling,
       ...config.json,
     },
+    use_multipart_params: config.use_multipart_params ?? false,
```
peter-leonov-ch: Do you think that this should be a config level option? I'd expect that a single client instance might be used to make different queries, some multipart and some not. Maybe, a query level override would be a good solution here. WDYT?

packages/client-node/__tests__/unit/node_multipart_query.test.ts
```diff
@@ -0,0 +1,215 @@
...
+      expect(options.headers['Content-Type']).toMatch(
+        /^multipart\/form-data; boundary=/,
+      )
+    })
+
+    it('should set Content-Type with a boundary when using multipart', async () => {
+      const adapter = buildHttpConnection({
+        compression: {
+          decompress_response: false,
+          compress_request: false,
```
peter-leonov-ch: Did you mean to add these parameters here? If not, please consider removing.

packages/client-web/src/connection/web_connection.ts
```diff
@@ -52,18 +53,40 @@ export class WebConnection implements Connection<ReadableStream> {
...
+    let body: string = params.query
+    let extraHeaders: Record<string, string> | undefined
+    if (useMultipart && params.query_params !== undefined) {
+      const boundary = `----clickhouse-js-${crypto.randomUUID()}`
+      body = buildMultipartBody({
+        query: params.query,
+        query_params: params.query_params,
+        boundary,
+      })
+      extraHeaders = {
```
peter-leonov-ch: I'd suggest that we follow the lead of the node counterpart here too and just extend existing headers instead of diverging with the code with introducing `extraHeaders`. I might be missing something though, WDYT?

peter-leonov-ch (commented):
Thanks a bunch for your contribution! This is a nice feature to have. Let's keep the conversation going 🚀

packages/client-common/src/utils/multipart.ts
jzila: The idea was that in the current version, the query is already sent in the body, with no boundary headers. We need the query to still be sent in the POST body, but now it must be separated from the query parameters, which are treated differently. Hence the boundary. If we don't boundary them off, then we need a naming convention to separate the form multipart keys that represent the query vs the keys that represent the query parameters.

jzila: Actually re-reading the server-side implementation, the boundaries seem unnecessary. The convention is purely naming based. `query` and `param_*`. I will remove the separate boundary for the query / param(s) content in the POST body.

peter-leonov-ch (dismissed):
This looks great, thanks for putting extra effort to address the comments. We'd need to add an integration test and add a line or two to the changelog. Would you want to do this part too or I can take it from here?

packages/client-common/src/utils/multipart.ts
```diff
@@ -0,0 +1,42 @@
...
+      `--${boundary}\r\n` +
+        `Content-Disposition: form-data; name="${name}"\r\n` +
+        `\r\n` +
+        `${value}\r\n`,
+    )
+  }
+
+  chunks.push(`--${boundary}--\r\n`)
+
+  return chunks.join('')
```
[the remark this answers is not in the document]
jzila: It constructs the payload using an Array.push, which is O(1). Then .join, which is O(n). So it's an extra O(n) operation compared to a buffered stream, but I think the simplicity of the code makes it worth it.

jzila:
@peter-leonov-ch Checking in here...this keeps coming up as needed in our use cases.

peter-leonov-ch:
@copilot resolve the merge conflicts in this pull request

peter-leonov-ch:
Hi @jzila — thanks for this! This PR has fallen out of date with `main` (it now conflicts after the repo-wide Prettier reformat and the `node_base_connection` socket-pool refactor landed).

I've rebased your four commits onto current `main` (authorship preserved, formatting normalized to the new Prettier config) and verified the common + Node.js unit suites pass locally. To get it merged, either option works:

1. **Enable "Allow edits from maintainers"** on this PR (checkbox in the right-hand sidebar), and I'll push the rebased commits directly to your branch so this PR updates in place — keeping all the review history here. I tried to push but currently get a permission error on the fork.
2. Alternatively, I've opened #825 with the rebased commits crediting you as the author. We can continue there if you'd prefer not to grant edit access.

Let me know which you'd like — happy to go whichever route is easiest for you.

peter-leonov-ch:
Thanks again for this contribution, @jzila — it's a genuinely useful feature, and the implementation and test coverage were in great shape.

Rather than ask you to re-sync a conflicting branch, I've rebased your four commits onto current `main` (authorship preserved, formatting normalized to the repo's Prettier config) and opened #825 to carry it forward. CI will run the full verification there.

Closing this in favor of #825 — all credit for the work remains yours. 🙏

PR #603, opened 2026-03-08
Fix ResultSet.json() race condition on JSONEachRow streams

Onyx2406:
## Summary

Fix `ResultSet.json()` throwing "Stream has been already consumed" on the **first** call for `JSONEachRow` format due to a race condition between two `readableEnded` checks.

Fixes #575

## Root Cause

For streamable JSON formats (JSONEachRow, etc.), `json()` calls `stream()` internally:

1. `json()` checks `this._stream.readableEnded` → **false** ✓
2. `json()` calls `this.stream()` 
3. **Race window**: Stream receives all data and fires 'end' event → `readableEnded` becomes **true**
4. `stream()` checks `this._stream.readableEnded` → **true** ✗
5. Throws: `"Stream has been already consumed"` — on the first call!

This is triggered by fast/small responses (e.g. `SELECT number FROM numbers(1)`) where the response arrives fully before `json()` reaches the `stream()` call.

## Fix (44 insertions, 2 files)

**`packages/client-node/src/result_set.ts`:**
- Add `_consumed` boolean flag to `ResultSet` class
- Add `markAsConsumed()` private method that checks `_consumed || readableEnded` and sets `_consumed = true` synchronously
- `text()`, `json()`, `stream()` all call `markAsConsumed()` as their first operation
- Split `stream()` into public `stream()` (with check) and private `_streamImpl()` (without check)
- `json()` calls `_streamImpl()` after already calling `markAsConsumed()`, eliminating the double-check race

This matches the pattern already used by the **web client's** `ResultSet` which uses `isAlreadyConsumed` boolean.

**`packages/client-node/__tests__/unit/node_result_set.test.ts`:**
- Add regression test that yields to the event loop (`setImmediate`) before calling `json()`, simulating the race condition

## Test Plan

- [x] New unit test for the race condition scenario
- [ ] CI (full test suite)
- Existing consumption tests (`text()`, `json()`, `stream()` only-once) continue to pass since `markAsConsumed()` covers all the same cases


packages/client-node/src/result_set.ts
peter-leonov-ch, line 142: Calling `this.markAsConsumed()` at the very top likely breaks this exception semantics. Before the PR change it would throw and still have the stream ready to be consumed in other way. Could you please check this scenario>

packages/client-node/src/result_set.ts
```diff
@@ -91,23 +92,38 @@ export class ResultSet<
...
+    this.markAsConsumed()
     return (await getAsText(this._stream)).toString()
   }
 
   /** See {@link BaseResultSet.json}. */
   async json<T>(): Promise<ResultJSONType<T, Format>> {
-    if (this._stream.readableEnded) {
-      throw Error(streamAlreadyConsumedMessage)
-    }
+    this.markAsConsumed()
```
peter-leonov-ch: Not in order. This line might better go into the `if`s below where we actually set the stream for consumption and an exception can only leave it in a potentially broken state. Still, it would be nice to check what happens if, say, one chunk has some sort of broken JSON but the next one in the stream is fine and the app tolerates broken chunks and allow for recover. We'd need to check this to avoid breaking long streaming clients that offload partial retries to a separate thread or similar.

packages/client-node/src/result_set.ts
```diff
@@ -126,14 +142,19 @@ export class ResultSet<
...
   /** See {@link BaseResultSet.stream}. */
   stream<T>(): ResultStream<Format, StreamReadable<Row<T, Format>[]>> {
-    // If the underlying stream has already ended by calling `text` or `json`,
-    // Stream.pipeline will create a new empty stream
-    // but without "readableEnded" flag set to true
-    if (this._stream.readableEnded) {
-      throw Error(streamAlreadyConsumedMessage)
-    }
-
+    // Validate format before marking as consumed, so that if the format is
```
peter-leonov-ch, line 147: This change can be avoided by making the consumed state change a tail update rather than head. Let's see if we can minimize the subtle API changes here.

peter-leonov-ch (commented):
Thanks a bunch for looking into this tricky issue and bringing a fresh look at the problem. Let's see if we can avoid making "dark" API changes here while addressing the issue in question.

peter-leonov-ch:
@Onyx2406, hey, thanks for waiting for the review. I think still can see some little issues with the explicit approach. But. Let's merge this and run full integration cycle to check the approach. I might tweak the solution a bit, hope it's ok with you. Again, thanks for waiting and for the excellent bug report!

peter-leonov-ch (approved):

PR #810, opened 2026-06-04
Use default prettier settings

peter-leonov-ch:
## Summary

Stick to the industry defaults as LLMs produce the code. Otherwise any code that comes in needs to be reformatted, especially subtrees from other repos.


mshustov:
I’m a bit concerned about the number of conflicts that need to be resolved for all the open PRs. What is the plan for these?

peter-leonov-ch:
> I’m a bit concerned about the number of conflicts that need to be resolved for all the open PRs. What is the plan for these?

Valid concern. I'm merging what's was about to be merged, the rest is going to be trivial to rebase with copilot. Our days of prettier rebase horror are finally gone 😎

PR #882, opened 2026-06-23
ci: add 7-day lockfile age audit for package-lock.json

motsc:
fails PRs that add `package-lock.json` entries published less than 7 days ago. complements the Dependabot cooldown by catching anything that bypasses it (manual edits, `--package-lock-only`, escape flags), audits the lockfile diff so it isn't tool-specific.

set-diff against `git merge-base origin/main HEAD`, not `+` lines, so lockfile reorders don't false-positive. existing lockfile grandfathered. ~1 registry request per added package, concurrency 8. `@clickhouse/*` excluded.

escape hatch is the `lockfile-age-skip` label, for legit CVE bumps inside the window.

failure looks like:

```
✗ 1 entries younger than 7 days:
  foo@1.2.3
    published: 2026-XX-XXT14:00:00Z (2 days ago)
    mergeable after: 2026-XX-XXT14:00:00Z
```

advisory only until repo admin adds `Lockfile age audit / audit` as a required check on main. will follow up once this lands.

companion: #881.

### DIFF
--- a/scripts/ci/lockfile-age-audit.mjs
+++ b/scripts/ci/lockfile-age-audit.mjs
@@ -130,7 +130,9 @@ const violations = []
 const errors = []
 
 async function checkOne({ name, version }) {
-  const url = `https://registry.npmjs.org/${name.replace('/', '%2F')}`
+  // Scoped names contain '/' which must be percent-encoded for the registry URL.
+  // encodeURIComponent handles all unsafe chars (replace('/', ...) only hits the first).
+  const url = `https://registry.npmjs.org/${encodeURIComponent(name)}`
   let res
   try {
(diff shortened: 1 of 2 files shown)


scripts/ci/lockfile-age-audit.mjs
```diff
@@ -0,0 +1,208 @@
...
+
+const queue = [...added.values()]
+const CONCURRENCY = 8
+async function worker() {
+  while (queue.length) {
+    const next = queue.shift()
+    if (next) await checkOne(next)
+  }
+}
+await Promise.all(Array.from({ length: CONCURRENCY }, worker))
```
peter-leonov-ch, line 181: this is likely gonna get us rate limited

PR #893, opened 2026-06-26
Standalone datatype-parser as a workspace package + RowBinary header compiler prototype

peter-leonov-ch:
## Summary

Brings the standalone ClickHouse data-type parser into the npm workspace and prototypes a compiler that turns a `RowBinaryWithNamesAndTypes` header into a stream reader.

### Workspace integration
- Link `@clickhouse/datatype-parser` into the npm workspace (shared root install/lockfile, `lint`/`typecheck`/`build` scripts, eslint config matching the other packages).
- Relocate the TS parser to `packages/datatype-parser` alongside the other packages; the C++ source it was ported from now lives at `packages/datatype-parser/reference-cpp-extracted-parser`.
- Fix path references (README link, `repository.directory`) and exclude the C++ reference dir from ESLint (its CMake `build/` tree emits `compiler_depend.ts` files).

### RowBinary header compiler (prototype)
- New `src/compile.ts` in the rowbinary skill: read the `RowBinaryWithNamesAndTypes` header, parse each column type string into an AST via `@clickhouse/datatype-parser`, **fold the AST into a tree of combinator calls** (`readArray(readNullable(readUInt32))`, `readMap(...)`, `readTupleNamed(...)`, …), and return a reader for the rest of the stream.
- No codegen/monomorphization yet — just the faithful AST → combinator fold, mirroring `readDynamicType`'s type→Reader mapping.
- Offline tests with a tiny `RowBinaryWithNamesAndTypes` encoder (no server needed).

### Notes
- The parser is wired into the skill as a **local `file:` dependency** (prototype-grade — the parser is not yet published to npm; this can't publish as-is).
- Variant alternative ordering relies on the server normalizing alternatives to sorted-type-name order in the header type string.


packages/datatype-parser/src/json.ts
```diff
@@ -11,7 +11,7 @@
 /// the `\u%04x` fallback). All bytes >= 0x20 are pushed through verbatim. At
 /// the end we decode the buffer back to a (byte-identical) JS string.
 
-import { NodeKind, type Node } from "./ast.js";
+import { NodeKind, type Node } from "./ast.ts";
 
 const encoder = new TextEncoder();
```
peter-leonov-ch, line 27: `npm run build --workspace @clickhouse/datatype-parser` currently fails here because this package tsconfig only includes `lib: ["ES2022"]`, so `TextEncoder`/`TextDecoder` are not in the type environment. Repro: `npx tsc -p packages/datatype-parser/tsconfig.build.json --noEmit` -> `TS2304`. Add the appropriate lib/types or import them from `node:util` before publishing.

skills/clickhouse-js-node-rowbinary-parser/src/rowBinaryWithNamesAndTypes.ts
```diff
@@ -0,0 +1,139 @@
...
+ */
+export function compileRowBinaryWithNamesAndTypes(
+  state: Cursor,
+  resolveType: TypeReaderResolver = typeStringToReader,
+): CompiledStream {
+  const { names, types } = readHeader(state);
+  const columnReaders = types.map((t) => resolveType(t));
+
+  const fields: Record<string, Reader<unknown>> = {};
+  for (let i = 0; i < names.length; i++) fields[names[i]!] = columnReaders[i]!;
```
peter-leonov-ch: Using a `Record` here loses wire order for legal headers: duplicate column names overwrite earlier columns, and integer-like names are reordered by `Object.keys()` inside `readTupleNamed`. RowBinary has no row delimiter, so reading fewer or out-of-order columns corrupts the stream (for example, two `AS x` UInt8 columns decode as multiple one-column rows). Keep an ordered reader list and handle name collisions explicitly when building the row object.

packages/client-node/src/index.ts
```diff
@@ -108,7 +108,9 @@ export const StreamableFormats = StreamableFormats_;
 export const StreamableJSONFormats = StreamableJSONFormats_;
 export const SingleDocumentJSONFormats = SingleDocumentJSONFormats_;
 export const RecordsJSONFormats = RecordsJSONFormats_;
+/** @deprecated Superseded by the `@clickhouse/datatype-parser` package (`parseDataType` + its `Node` AST); slated for removal in a future major version. */
```
peter-leonov-ch, line 111: This public deprecation guidance plus the new publishable `@clickhouse/datatype-parser` package is an observable API/distribution change, but `CHANGELOG.md` has no `1.23.0` entry for it. AGENTS.md requires changelog updates in the same PR for public API/behavior changes.

skills/clickhouse-js-node-rowbinary-parser/package-lock.json
```diff
@@ -8,10 +8,25 @@
...
         "typescript": "^6.0.3",
         "vitest": "^4.1.9"
+      },
+      "engines": {
+        "node": ">=20"
+      }
+    },
+    "node_modules/@clickhouse/datatype-parser": {
+      "version": "0.1.1",
+      "resolved": "https://registry.npmjs.org/@clickhouse/datatype-parser/-/datatype-parser-0.1.1.tgz",
```
peter-leonov-ch, line 25: The skill lockfile resolves `@clickhouse/datatype-parser` from npm, so the RowBinary CI job that runs `npm ci` inside `skills/clickhouse-js-node-rowbinary-parser` does not exercise the local `packages/datatype-parser` changes in this PR. A local parser regression can pass CI against the previously published parser. Use a workspace install, local packed tarball, or otherwise force this dependency to resolve to the checkout.

packages/datatype-parser/package.json
```diff
@@ -1,20 +1,17 @@
...
-      "types": "./dist/src/index.d.ts",
-      "import": "./dist/src/index.js"
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js"
     }
   },
-  "bin": {
-    "chdt-parse": "dist/tool/main.js"
-  },
   "files": [
```
peter-leonov-ch, line 15: `origin/main` exposed a `bin` entry for `chdt-parse` and built `tool/**/*.ts`; this package now publishes only `dist` from `src` and has no `bin`, while the README still documents the CLI. Unless dropping the CLI is intentional, keep the `bin` entry and include/build the tool output.

packages/datatype-parser/test/parser.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  it("deliberately-unsupported types are rejected with a hard error", () => {
+    for (const typeStr of readCases(join(here, "cases_unsupported.txt"))) {
+      const r = parseDataType(typeStr);
+      expect(r.ok(), `expected ${typeStr} to be rejected`).toBe(false);
+      expect(r.error && r.error.message.length > 0).toBeTruthy();
+    }
+  });
+
+  it("trailing input after a complete type is an error", () => {
```
peter-leonov-ch, line 130: Please add a malformed-number regression here. The lexer currently accepts an exponent marker/sign without requiring at least one exponent digit, so `Decimal(1e, 2)` parses successfully. `toJSON` then emits invalid JSON (`"value":1e`) because Float64 literals are written raw. The parser should reject malformed exponents and the JSON path should be covered.

packages/datatype-parser/src/json.ts
peter-leonov-ch: Fixed in 7536302c. Declared `TextEncoder`/`TextDecoder` locally in `json.ts` (just the slice we use) instead of adding `@types/node` or the `DOM` lib — keeps the package dependency-free and avoids leaking DOM globals (`window`, `name`, …) into every file. `tsc -p tsconfig.build.json --noEmit` now passes.

skills/clickhouse-js-node-rowbinary-parser/src/rowBinaryWithNamesAndTypes.ts
peter-leonov-ch: Fixed in 7536302c. The row reader now reads POSITIONALLY by column index rather than building a name-keyed `readTupleNamed`. Duplicate names no longer collapse (every column is still read off the wire) and integer-like names no longer get reordered by `Object.keys()` — reads always follow header order, so the cursor stays in sync. Last-name-wins now only affects the output object, not the framing. Added an offline regression test (`tests/rowBinaryWithNamesAndTypes.order.test.ts`) covering both the duplicate-name and integer-like-name headers.

packages/client-node/src/index.ts
peter-leonov-ch: Fixed in 7536302c. Added `1.23.0` entries: a Migration Note for the `parseColumnType` / `SimpleColumnTypes` deprecation and a New-features entry for publishing `@clickhouse/datatype-parser`, plus the `[#893]` link ref. Ran `prettier --write CHANGELOG.md`.

skills/clickhouse-js-node-rowbinary-parser/package-lock.json
peter-leonov-ch: Fixed in 7536302c. The skill CI now builds + packs the in-repo `@clickhouse/datatype-parser` and installs it over the registry version with `npm install <tarball> --no-save` (the committed `^0.1.1` dep and lockfile stay intact for end users, only `node_modules` is overridden). `packages/datatype-parser/**` is now also a path trigger for this workflow, so a local parser regression can no longer pass this suite against the previously published parser.

packages/datatype-parser/package.json
peter-leonov-ch: Fixed in 7536302c. Dropping the CLI is intentional (the build is library-only), so cleaned it up rather than restoring `bin`: removed `tool/main.ts`, the `parse` script, the `tool/**` tsconfig include, and the README Layout row. The package now publishes only the library (`dist` built from `src`), consistent with `tsconfig.build.json`.

packages/datatype-parser/test/parser.test.ts
peter-leonov-ch: Fixed in 7536302c. The lexer now requires at least one digit after `e`/`E[+/-]`, so `Decimal(1e, 2)` and `Decimal(1e+, 2)` are rejected with `malformed number: exponent has no digits` — the invalid `"value":1e` JSON can no longer be produced. Added a regression test asserting rejection of the malformed forms and a valid-JSON round-trip for a well-formed exponent (`Decimal(1e3, 2)`).

packages/datatype-parser/src/parser.ts
```diff
@@ -62,6 +62,17 @@ function isEnumTypeUpper(u: string): boolean {
...
 }
 
+/// Does `text` parse in FULL as a numeric literal — the TS equivalent of the
+/// C++ `strtod` (float) / `strtoull` (integer) + end-pointer check? The lexer
+/// is deliberately lenient and can hand back a half-formed lexeme such as `1e`
+/// (an exponent with no digits); those must be rejected. Integer lexemes the
+/// lexer emits are pure digit runs (so always valid); floats may carry a
+/// malformed exponent, so the optional exponent here requires >= 1 digit.
+function isFullyParsedNumber(text: string, isFloat: boolean): boolean {
+  if (isFloat) return /^(?:\d+\.?\d*|\.\d+)(?:[eE][+-]?\d+)?$/.test(text);
```
peter-leonov-ch, line 72: Malformed exponent handling is fixed, but this still accepts float spellings that are not JSON numbers. For example, `parseDataType('Decimal(1., 2)')` and `parseDataType('Decimal(.5, 2)')` return `ok=true`; then `toJSON` writes `"value": 1.` / `"value": .5` verbatim, so `JSON.parse(toJSON(ast))` throws. Since `toJSON` is public API and `json.ts` assumes Float64 literals are already valid JSON numbers, please either reject these lexemes here or normalize Float64 output before writing it raw.

packages/datatype-parser/package.json
```diff
@@ -1,20 +1,17 @@
 {
   "name": "@clickhouse/datatype-parser",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "description": "Standalone ClickHouse data-type string parser — a TypeScript port of the chdt C++ library.",
   "license": "Apache-2.0",
   "type": "module",
-  "main": "dist/src/index.js",
-  "types": "dist/src/index.d.ts",
+  "main": "dist/index.js",
```
peter-leonov-ch, line 7: `chdt-parse` still is not included in the published package. The previous package had `bin: { chdt-parse: dist/tool/main.js }` and built `tool/**/*.ts`; this moved package has no `bin` entry, `tsconfig.build.json` emits only `src`, and the TS `tool/main.ts` is deleted. That removes `npx chdt-parse` for consumers of `@clickhouse/datatype-parser`. If that removal is intentional it needs to be called out as a breaking/package change; otherwise please restore the bin and tool build.

packages/datatype-parser/src/parser.ts
peter-leonov-ch: by design

packages/datatype-parser/package.json
peter-leonov-ch: removed completely

PR #909, opened 2026-06-28
test(upstream): expand rowbinary allowlist after float/datetime fixes

peter-leonov-ch:
## What

Re-ran the differential sweep over the full passthrough `upstream-allowlist.txt` with the Float32 / float-text / DateTime64 fixes from #908 merged, and expanded `rowbinary-allowlist.txt` with the tests that now decode-match passthrough byte-for-byte: **2239 → 2264**. Union with the prior list, so it can only add tests — no regressions.

## Scope note (honest expectation)

The realized gain is **modest** (~+25). Most of the ~600 passthrough tests the rowbinary backend doesn't yet cover are **not** Float32 mismatches — they need **renderer features that don't exist yet** (`Variant`, `JSON`, `Dynamic`, geo, `Nested`) or are **non-deterministic** (can't be validated against a static `.reference`). Closing that remainder is a separate, larger effort (new type renderers), not this PR.

## On validation

A local standalone sweep is **not** a reliable oracle here: some tests produce different (or empty) output outside the upstream Python `clickhouse-test` runner, so `rowbinary == passthrough` can hold trivially standalone yet say nothing about the `.reference` diff. The authoritative check is the **CI `rowbinary` matrix leg** (Python runner vs the static `.reference`) — which is green on this PR.

### DIFF
--- a/.github/workflows/upstream-sql-tests.yml
+++ b/.github/workflows/upstream-sql-tests.yml
@@ -6,9 +6,9 @@ on:
   workflow_dispatch:
     inputs:
       upstream_ref:
-        description: "ClickHouse/ClickHouse ref to check out"
+        description: "ClickHouse/ClickHouse ref to check out (defaults to the pinned commit)"
         required: false
-        default: "master"
+        default: ""
         type: string
   schedule:
     - cron: "0 5 * * *"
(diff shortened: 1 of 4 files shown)


peter-leonov-ch (commented):
LGTM

peter-leonov-ch (commented):
LGTM

Issue #910, opened 2026-06-28
readVariant should return the active type discriminant, not just the value


peter-leonov-ch:
## Summary

`readVariant` in `@clickhouse/rowbinary` (`skills/clickhouse-js-node-rowbinary-parser/src/composite.ts`) reads the 1-byte discriminant to pick which alternative reader to run, then returns **only the decoded value** and discards the discriminant. For most `Variant`s the active alternative can be inferred from the JS value's shape, but for some it cannot — two alternatives can decode to an identical JS value yet correspond to different ClickHouse types, so a consumer cannot tell which one it got.

## Why it matters

A consumer (e.g. a text serializer, or any user inspecting decoded values) needs the active type to format/interpret the value correctly. Verified against a live server (26.1):

| Variant | value | decoded JS (both alternatives) | correct ClickHouse text |
| --- | --- | --- | --- |
| `Variant(UInt8, Enum8('a'=1,'b'=2))` | `1` | `number 1` (identical) | `1` (UInt8) vs `a` (Enum8) |
| `Variant(Date, DateTime)` | midnight | `Date(…T00:00:00)` (identical) | `2020-01-02` vs `2020-01-02 00:00:00` |

In both rows the decoded value is indistinguishable, so without the discriminant there is no way to recover the right interpretation. (Cases like `Variant(String, UInt64)` or `Variant(Int32, Int64)` are unaffected — the shapes differ or the text is identical.)

These specific collision patterns (a numeric type beside an `Enum`, or `Date` beside `DateTime`/`DateTime64`) are **rare** — they don't appear in any of the upstream `0_stateless` tests that use `Variant` — so this is a **correctness/robustness** fix for arbitrary user input, not a coverage blocker.

## Proposed change

Have `readVariant` surface the active alternative index alongside the value, e.g.

```ts
// today
export function readVariant(readers): Reader<T[number] | null>
// proposed (shape TBD)
export function readVariant(readers): Reader<{ typeIndex: number; value: unknown } | null>
// (NULL discriminant 0xFF -> null, as today)
```

The reader already knows the index (it dispatches on the discriminant), so this is just threading it through. Consumers that don't care can ignore `typeIndex`.

## Notes / considerations

- **Breaking change** to `readVariant`'s return type — should land with a minor/major bump and a changelog note.
- The type-AST → reader fold (`astToReader` in `compile.ts`) maps the discriminant index to `node.arguments[typeIndex]`; a renderer can then format with the exact subtype.
[...]

PR #912, opened 2026-06-28
feat(rowbinary)!: resolve Enum8/Enum16 to their names

peter-leonov-ch:
## What

`readEnum8` / `readEnum16` previously returned the raw underlying `Int8`/`Int16`; the name had to be resolved by the caller. But people reach for enums precisely for the *label*, so this makes name resolution the default.

They are now **factories** that take the enum's `value -> name` map and return a `Reader<string>` yielding the name:

```ts
// before
readEnum8(state: Cursor): number
// after
readEnum8(valueToName: EnumNameMap): Reader<string>   // EnumNameMap = ReadonlyMap<number, string>
```

Both decode paths build the map from the `'name' = value` pairs the type already carries and pass it in:
- `compile.ts` (textual `RowBinaryWithNamesAndTypes` header) — from `EnumDataType.values`.
- `dynamic.ts` (binary type encoding, tags `0x17`/`0x18`) — from the name/value pairs in the header (previously read and discarded).

An unmapped value falls back to its stringified integer, so a decode never throws on an unexpected wire value. The **raw integer** is still a first-class path via `readInt8`/`readInt16` — the example readers (`orders`, `observability`) use those to keep their monomorphized hot paths allocation-free.

## Breaking change

`readEnum8`/`readEnum16` changed signature, and decoded enum values are now `string` (name) rather than `number`. Pre-1.0 package, so a minor bump should carry it; flagged with a `BREAKING CHANGE:` footer.

## Tests

- Rewrote `Enum8`/`Enum16` unit tests to assert name resolution + the unmapped-fallback, and updated the `framing`, `rowBinaryWithNamesAndTypes` (header path), and `Dynamic` (binary path) enum assertions.
- Full skill suite green: **564 passing** (against a live CH 26.1), `tsc --noEmit` clean, prettier clean.

## Heads-up for the test-runner harness (separate package, not in this PR)

`tests/clickhouse-test-runner`'s TSV serializer maps enum int→name itself (`enumName`) and currently expects a `number`. It consumes the **published** `@clickhouse/rowbinary`, so it's unaffected until this is released and the dep is bumped — at which point that serializer's enum case must adapt to receive the name directly.

Context: split out from #909 review; complements the Variant discriminant discussion in #910.


peter-leonov-ch (commented):
LGTM

PR #998, opened 2026-08-14
fix(types): re-export public API types dropped in 1.23.0

### LINKED ISSUE #997: Some public API types are not exported from @clickhouse/client since 1.23.0
**Describe the bug**

Since 1.23.0, when `@clickhouse/client-common` was bundled into the client (#845), a few types that the client's own public methods use are not re-exported from `@clickhouse/client`:

- `QueryParamsWithFormat`: the parameter of `query()`
- `PingParams`: the parameter of `ping()`
- `ResultJSONType` and `RowJSONType`: what `ResultSet.json()` and `Row.json()` return
- `ClickHouseSummary` and `WithClickHouseSummary`: fields of `InsertResult` and `CommandResult`

I could not find a supported place to import them from. `@clickhouse/client-common` is deprecated, and its npm deprecation message says to import from `@clickhouse/client` instead, but these types are not there. There is also no `exports` map in the package, so the only options left are copying the type into my own code or importing from `dist/`.

The internal things (`Connection*`, `LogWriter`, `getConnectionParams` and so on) are obviously left out on purpose, but these ones are part of the public API. Probably the same miss as #935, where `EXCEPTION_TAG_HEADER_NAME` and `extractErrorAtTheEndOfChunk` were added back in 1.23.1.

**Steps to reproduce**

1. `npm i @clickhouse/client@1.23.1 typescript`
2. Put the code below into `probe.ts`
3. `npx tsc --noEmit probe.ts --module nodenext --moduleResolution nodenext --target esnext --strict`

**Expected behaviour**

If a type is used in the signature of a public method, it should be importable from the package root. Otherwise you cannot write a wrapper with the same signature.

**Code example**

```ts
import type {
  ClickHouseSummary,
  PingParams,
  QueryParamsWithFormat,
  ResultJSONType,
  RowJSONType,
  WithClickHouseSummary
} from '@clickhouse/client'
```

What I needed it for: a wrapper around `query()` that adds a `query_id` to every request, so the query can be killed on abort. To keep format inference it has to take the same params as `query()`:

```ts
async function chQuery<Format extends DataFormat = 'JSON'>(
  params: QueryParamsWithFormat<Format> // cannot be imported
) { /* ... */ }
```

Copying the type from `common/client.d.ts` as is works fine, so I don't think anything changed in the API here:

```ts
type QueryParamsWithFormat<Format extends DataFormat> = Omit<QueryParams, 'format'> & {
  format?: Format
}
```

**Error log**

```
probe.ts(2,3): error TS2305: Module '"@clickhouse/client"' has no exported member 'ClickHouseSummary'.
probe.ts(3,3): error TS2305: Module '"@clickhouse/client"' has no exported member 'PingParams'.
probe.ts(4,3): error TS2305: Module '"@clickhouse/client"' has no exported member 'QueryParamsWithFormat'.
probe.ts(5,3): error TS2305: Module '"@clickhouse/client"' has no exported member 'ResultJSONType'.
probe.ts(6,3): error TS2305: Module '"@clickhouse/client"' has no exported member 'RowJSONType'.
probe.ts(7,3): error TS2305: Module '"@clickhouse/client"' has no exported member 'WithClickHouseSummary'.
```

**Configuration**

- Client version: 1.23.1, same on `1.24.0-head.443b2f0.1` and on `@clickhouse/client-web@1.23.1`
- Node.js 26.7.0, TypeScript 7.0.2
- macOS

Saitgalin:
Closes #997. The issue lists six of these; I found the other four (the ping ones and `ResultStream`) while preparing the change.

All ten were exported from the root of `@clickhouse/client-common`, and the ping ones sit in the block that its `index.ts` marks with `/** Should be re-exported by the implementation */`. The lower level exports (`Connection*`, `LogWriter`, the encoders and utils) are clearly internal, so I left them out. `WithHttpStatusCode` was never exported from `client-common` either, so it is not here; happy to add it if you want the whole intersection nameable.

Type-only re-exports: I built the tree with and without the change, `dist/index.js` is byte identical for both packages, only the `.d.ts` files differ. Locally green: typecheck, lint, prettier, build, `test:node:unit` (408 passed). `test:web:unit` needs Playwright browsers, which I don't have set up.

No test included, and I would rather ask than guess: a type level assertion would not fail anywhere in CI today, since the package tsconfigs include only `src/**` and vitest runs without `typecheck`.


mshustov (approved):

mshustov:
@Saitgalin could you sign CLA pls?

Saitgalin:
Signed. CLA check is green now.

